PR #22, opened 2025-06-05
Allow custom tenant ID and enterprise app

2underscores:
Adding ability to set own enterprise app client ID in own tenant. 

Better security and control giving ability to only use your own enterprise apps. Still defaults to original one.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -17,8 +17,8 @@ const FALLBACK_PATH = path.join(FALLBACK_DIR, '..', '.token-cache.json');
 
 const DEFAULT_CONFIG: Configuration = {
   auth: {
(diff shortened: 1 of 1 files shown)


eirikb:
I've been contemplating this, but landed on that I wanted the hard coded app to be used. It is completely safe.  
The app itself has no permissions, all of the permissions come from the code, without ability for automatic escalation.  
It should be impossible to indirectly get hold of any data - unless via the MCP server itself, and well then another app registration won't help.  
The code is OSS so that won't happen, and the npm package is published from GitHub. All good. All safe.

But I've seen others fork just for this, and it would be easier to set up on a business without having to consent to our app registration - so I'll add it :smile_cat: 

Thank you for the PR :slightly_smiling_face:

Issue #37, opened 2025-06-18
Oauth Authorization Code Flow support


virtuallifenl:
Hi,

The MCP standard asks of MCP Servers to implement certain standardized Oauth flows:
https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#authorization-flow

Can you perhaps implement Oauth discovery and authorization code flow support?

If this is not possible, I would like a way to just POST to the endpoint with a bearer token that I received with an oauth authorization flow outside of the MCP Server. The MCP Server can then just forward this bearer token to the Graph API.

eirikb:
Hi. The device code flow we already have is OAuth and for me at least it works very well.

If you mean to start up a web server and do more "normal" auth flow against localhost then that is possible, but IMHO it is neither better or safer. This was actually my initial plan, as I've used this approach before, but ended on code flow instead.

virtuallifenl:
I am not asking for the (more usual) Authorization Code Flow because I think it's safer or better. I would argue that using discovery and the Authorizatin Code Flow is what the MCP standard itself uses and therefore (MCP) clients are also implementing this standard. So MCP clients can easily use the discovery feature of OAuth to initiate and follow the flow. This server can't be used by clients that implement the MCP standard as written. Although I understand, of course, that the current flow will also work in some scenarios - adding the way the MCP standard describes it, would make it usable in even more scenarios. I hope you are willing to consider making this server even more widely usable with more choice for its users.

If implementing the standard is too much work, I would like it very much you could just "pass through" the Authorization header to the Graph API so that people can - optionally - implement authorization outside of this MCP server.

I hope you are willing to consider either of these options. Not because they are better or safer, but because they add options for scenario's with clients that strictly follow the MCP standard.

eirikb:
It shouldn't be too hard to add support.  
This looks interesting as well; https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#proxy-authorization-requests-upstream . That's the SDK I use.

eirikb:
> what does all this mean for multi user support? compared to this project https://github.com/jibberish69/enhanced-outlook-mcp which allows multiple users (i assume by passing on tokens)

Ok so this sever looks like it is using the deprecated SSE HTTP protocol, running on port 3000, while doing auth on localhost port 3333.
It seems this server creates a random id and assign this to the client, somehow.  
[...]

PR #67, opened 2025-07-02
Fixing proper auth Flow as per MCP + Oauth Spec

leondape:
This is basically a Fix for #37  and #43. 

## Status quo: 
- http://localhost:3000/authorize was implemented. 

## Missing for full flow with dynamic client registration:
- http://localhost:3000/authorize
- http://localhost:3000/register
- http://localhost:3000/token

### Test with inspector:
Bildschirmfoto 2025-07-02 um 10 10 11

It now works with the inspector. You just need some callback routes inside the m365 app registration.

### Test with other Clients 
There are not a lot clients out there which support this new MCP spec with Oauth. Claude Desktop for example does NOT!
You can try it out with https://cursor.com/. It recognized that login is required and will issue a login. However you will need to set cursors redirect URI in microsoft app registration again. 

Bildschirmfoto 2025-07-02 um 10 07 21

## Summary
- Introduced OAuth token handling in GraphClient, allowing for access and refresh tokens to be set and used for API requests.
- Added endpoints for OAuth authorization, token exchange, and dynamic client registration.
- Updated README with Azure AD setup instructions for OAuth testing.
- Enhanced server logging for better debugging and environment variable checks.
- Added CORS support and health check endpoint for the server.

## Important considerations before merging or using in production:‼️potentially Vibe-Coded-Slop‼️
- I did this with my poor understanding of the standard and am not a really experienced web-dev. Therefore please pleas please review all this well before merging it. It works, but I don't know if it is secure as well. 

There are some things I would consider as best practices to check before rolling it out, inspired by this [MCP-repo](https://github.com/LibreChat-AI/spotify-mcp?tab=readme-ov-file#not-production-ready):

1. Add Authentication Validation

- Implement proper client validation in the /register endpoint
- Add rate limiting to prevent abuse
- Validate redirect URIs against a whitelist

2. Enhance Security
- Implement CSRF protection
- Add request signing/verification
- Use secure session management
- Implement proper error handling that doesn't leak sensitive information

3. Add Monitoring & Logging

- Implement comprehensive logging
- Add error tracking (e.g., Sentry)
- Monitor API usage and rate limits
- Track OAuth flow completion rates

5. Handle Edge Cases
- Implement proper retry logic with exponential backoff
- Add graceful degradation for non-critical features
- Implement proper timeout handling

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,24 @@
+# Microsoft 365 OAuth Configuration
+# Create an Azure AD app registration and get these values:
+
+# Your Azure AD App Registration Client ID
+MS365_MCP_CLIENT_ID=your-azure-ad-app-client-id-here
+
+# Your Azure AD App Registration Client Secret
+MS365_MCP_CLIENT_SECRET=your-azure-ad-app-client-secret-here
+
+# Tenant ID - use "common" for multi-tenant or your specific tenant ID
+MS365_MCP_TENANT_ID=common
+
+# Instructions:
+# 1. Go to https://portal.azure.com
+# 2. Navigate to Azure Active Directory → App registrations → New registration
+# 3. Set name: "MS365 MCP Server"
+# 4. Add these redirect URIs (for MCP Inspector testing):
+#    - http://localhost:6274/oauth/callback
+#    - http://localhost:6274/oauth/callback/debug
+#    - http://localhost:3000/callback (optional, for server callback)
+# 5. Copy the Client ID from Overview page
+# 6. Go to Certificates & secrets → New client secret → Copy the secret value
+# 7. Replace the values above with your actual credentials
+# 8. Rename this file to .env 
--- a/README.md
+++ b/README.md
@@ -144,6 +144,35 @@ This mode:
 
 MCP clients will automatically handle the OAuth flow when they see the advertised capabilities.
 
+##### Setting up Azure AD for OAuth Testing
+
+To use OAuth mode with custom Azure credentials (recommended for production), you'll need to set up an Azure AD app registration:
+
+1. **Create Azure AD App Registration**:
+   - Go to [Azure Portal](https://portal.azure.com)
+   - Navigate to Azure Active Directory → App registrations → New registration
+   - Set name: "MS365 MCP Server"
+
+2. **Configure Redirect URIs**:
+   Add these redirect URIs for testing with MCP Inspector:
+   - `http://localhost:6274/oauth/callback`
+   - `http://localhost:6274/oauth/callback/debug`
+   - `http://localhost:3000/callback` (optional, for server callback)
+
+3. **Get Credentials**:
+   - Copy the **Application (client) ID** from Overview page
+   - Go to Certificates & secrets → New client secret → Copy the secret value
+
+4. **Configure Environment Variables**:
+   Create a `.env` file in your project root:
+   ```env
+   MS365_MCP_CLIENT_ID=your-azure-ad-app-client-id-here
+   MS365_MCP_CLIENT_SECRET=your-azure-ad-app-client-secret-here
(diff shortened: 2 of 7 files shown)


leondape:
About the security implementations: I have no idea! 
It's vibe-coded / copyied from other recent MCP Servers which implement the standard well. 

About PKCE and the use-case I might be able to shed some light there: 

## Usecase: 
Single Deployment - Multiple Users via a singular Chat Platform. The platform handles the multi-user part. All that is needed if you don't want to administrate multiple Apps on device but rather use a web-app platform. Also, all public MCP servers will work like this.  
For corporate environments this will mean that you only deploy one MCP server in the cloud and all other users can connect to it and use it with their our auth.

Adhering to the standard will make it always out of the box compliant with modern clients and deployable anywhere as "public" MCP Servers

## The Creds
I cannot tell you 100%. But as far as I understood the PKCE are not used for accessing anything. They are used for the client registration process of "dynamic client registration". 
1. MCP client wants to use the Server
2. requests connection
3. must auth at microsoft
4. if auth successful, client is allowed to use server 
5. now user wants to use the server, user needs separate tokens 
6. User will need to grant access once more 

For single users this is not needed. For multiple users this makes sense as you have one client level initialization and the a user level auth. 

No idea if that makes sense...

eirikb:
I kind of see the point of having a single server within a company, but some setup is needed (host it, app registration etc.)
And for clients such as claude code or claude.ai it actually has to be public.  
If someone were to connect to a public server, and it was set up as multi-tenant, it would probably work fine - while tokens are in theory available for that server. Not directly MITM but data is available with some effort. In my mind this is scary stuff.

I also find it strange that the whole flow thing isn't part of the SDK.

Oh well, now we have it. Thanks :slightly_smiling_face:

leondape:
**On the security model:**
You're right that this requires trust in the server operator, but that's actually the standard OAuth pattern for any multi-user service (think Slack, GitHub, etc.). The key security aspects:

1. **Server never sees raw user credentials** - only OAuth tokens from Microsoft
2. **Tokens are scoped** to specific permissions the user explicitly grants
3. **Tokens can be revoked** by the user at any time via Microsoft's portal
4. **Server operator can't impersonate users** - they'd need the user's Microsoft password/MFA

The "data available with some effort" concern is real, but it's the same trust model as any SaaS that integrates with Microsoft 365. The alternative (every user running their own server) just isn't practical for most corporate deployments.

**Why not pure MSAL/SDK:**
Great question! The issue is we need to act as **both**:
- OAuth client (to Microsoft) - this could use MSAL
- OAuth server (for MCP clients) - this needs custom implementation

The MCP SDK's OAuth features are designed for generic providers, but Microsoft has specific requirements (tenant handling, Graph API validation, refresh flows) that need custom code. We're essentially creating a "Microsoft-aware OAuth proxy."

**PKCE vs secrets:**
You're absolutely right that PKCE should be sufficient for public clients! The client secret requirement comes from Microsoft's token exchange endpoint when we're acting as a confidential client. In theory we could make this work with PKCE-only, but it would require registering as a public client with Microsoft, which has different security characteristics.

The current implementation follows the "authorization code with client credentials" pattern that most enterprise OAuth integrations use. It's more complex but gives better audit trails and admin control.

**SDK limitations:**
Totally agree this should be in the SDK! The current MCP SDK OAuth implementation is fairly basic. Ideally we'd have provider-specific packages like `@mcp/auth-microsoft`, `@mcp/auth-google`, etc. that handle these integration patterns cleanly.

For now, this gives us a working multi-user solution that follows OAuth 2.1 best practices, even if the implementation is a bit verbose.

leondape:
Understood. In an ideal world I am waiting for microsoft to host an MCP server themselves. 
The nice thing is that any company can host a relay station for themselves in their perimeter.

PR #115, opened 2025-08-18
feat: Implement client credential authentication flow

ariavn-byte:
This commit implements the client credential authentication flow to allow the server to authenticate non-interactively using a client ID and client secret.

Key changes:
- The `AuthManager` in `src/auth.ts` now detects the presence of the `MS365_MCP_CLIENT_SECRET` environment variable.
- When the secret is present, it uses `ConfidentialClientApplication` to acquire a token via the client credential flow. This bypasses the interactive device code flow.
- When the secret is not present, it retains the existing `PublicClientApplication` for interactive logins.
- Hardcoded fallback Client IDs have been removed from `src/server.ts` and `src/auth.ts` to ensure that only user-provided credentials are used.
- A `start` script has been added to `package.json` to standardize the production startup command for deployment platforms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "generate": "node bin/generate-graph-client.mjs",
     "postinstall": "npm run generate",
     "build": "tsup",
+    "start": "node dist/index.js --http ${PORT-3000}",
     "test": "vitest run",
     "test:watch": "vitest",
     "dev": "tsx src/index.ts",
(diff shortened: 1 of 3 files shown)


eirikb:
Hi @ariavn-byte . If you fix the things I mentioned - and by that I mean actually fix them, not only tag them as resolved without resolving them, then I will accept the PR.

PR #116, opened 2025-08-22
Fix calendarId parameter handling for calendar operations

cvsloane:
## Summary

This PR fixes an issue where the calendarId parameter was not being properly handled for calendar operations, causing all events to be created on the default calendar instead of the specified calendar.

## Changes Made

- Enhanced calendar operation detection: Added logic to identify calendar-specific operations
- Dynamic path modification: When calendarId is provided, the API path is dynamically changed from /me/events to /me/calendars/{calendarId}/events
- Parameter filtering: Added proper handling to skip calendarId parameter from being passed as a query parameter since it's now used in the path
- Comprehensive operation support: The fix works for all calendar operations, not just event creation
- Backwards compatibility: When no calendarId is provided, operations default to the original /me/events endpoint

## Technical Details

The core issue was in src/graph-tools.ts where the endpoint paths were hardcoded. The fix adds special handling around line 147-167 to intercept calendarId parameter and modify the API path dynamically.

## Testing

- Added unit tests in test/calendar-fix.test.js to verify the logic
- Tested with real Microsoft 365 calendars using test-real-calendar.js
- Verified events are created on the correct calendar when calendarId is specified
- Confirmed backwards compatibility when calendarId is not provided

## Resolves

This fixes the issue where users could not create calendar events on specific calendars other than their default calendar, which is essential for multi-calendar management scenarios.


eirikb:
Hi there. Are you sure we should have tool-specific handling of this? Perhaps the id should be encoded for every path?

cvsloane:
You make an excellent point! You're absolutely right that a more general solution would be better than tool-specific handling.

Looking at the Microsoft Graph API patterns, many endpoints follow similar structures:
- `/me/events` vs `/me/calendars/{calendarId}/events`
- `/me/messages` vs `/me/mailFolders/{folderId}/messages`
- etc.

A more elegant approach would be to:
1. Define path parameters in the endpoint definitions that can modify the base path
2. Automatically encode any ID parameters (like `calendarId`, `folderId`, etc.) into the path structure
3. Remove the need for tool-specific logic

Would you prefer I refactor this to use a more general parameter-to-path encoding system? I could update the approach to handle ID parameters generically rather than having calendar-specific code.

The current fix works and solves the immediate problem, but I agree your suggested approach would be more maintainable long-term. Let me know if you'd like me to rework this with a more general solution!

eirikb:
Let's not add the .md-files

PR #117, opened 2025-08-22
Harden server for production deployment

ariavn-byte:
This commit prepares the server for a production deployment on platforms like Render.

The following changes were made:

- A `start` script was added to `package.json` to provide a standard way to run the application in production with HTTP mode enabled.

- The server is updated to respect the `PORT` environment variable, which is a common requirement for cloud deployment platforms.

- A startup check is added to `src/server.ts`. The server will now exit with an error if it is started in HTTP mode without the required `MS365_MCP_CLIENT_ID` and `MS365_MCP_CLIENT_SECRET` environment variables. This prevents running in a misconfigured state.

- The `README.md` is updated with a new "Deployment" section that provides clear instructions on how to deploy the server, including the necessary build/start commands and environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,28 @@ Or configure Claude Desktop manually:
 
 > **Note**: Run `npm run build` after code changes to update the `dist/` folder.
 
+## Deployment
+
+This server is designed to be deployed to cloud platforms like [Render](https://render.com/).
+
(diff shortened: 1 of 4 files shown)


eirikb:
Sorry but I'm closing this now

PR #118, opened 2025-08-25
Fix/package json syntax

ariavn-byte:


ariavn-byte (commented):
appoved

eirikb:
Why are you setting keytar as optional?

eirikb:
Ok, so I had a second look, and I think having it optional fits you quite well, but it won't fit other users that use this server with default setup. The default setup, with stdio, is completely dependent on keytar - or at least it wouldn't be nice to live with.

However, I think it could be nice for HTTP users, but then we need a message to stdio users where keytar doesn't load. Agree?

eirikb:
Hi @ariavn-byte . The keytar change is fine - the rest I don't know. We already have npm so I don't think we need GitHub registry? And the other things... I just don't know. Perhaps for this PR stick to the optional keytar thing? If you revert the rest then I will accept the PR.

PR #127, opened 2025-09-13
fix: correct https detection and registration

mfa326:
Ensured OAuth discovery endpoints use the correct HTTPS scheme by trusting proxy headers and deriving base URLs from x-forwarded-proto and x-forwarded-host, preventing http origins on TLS-terminated deployments

Added dynamic client registration metadata (client_id_issued_at and client_secret_expires_at) and exposed issuer configuration via MS365_MCP_ISSUER_URL for OpenAI MCP compliance

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -92,6 +92,15 @@ class MicrosoftGraphServer {
       const app = express();
       app.use(express.json());
       app.use(express.urlencoded({ extended: true }));
+      // Support TLS termination proxies
+      app.set('trust proxy', true);
+
+      const getBaseUrl = (req: Request): URL => {
+        const proto =
+          (req.headers['x-forwarded-proto'] as string)?.split(',')[0] || req.protocol;
+        const host = (req.headers['x-forwarded-host'] as string) || req.get('host');
+        return new URL(`${proto}://${host}`);
+      };
 
       // Add CORS headers for all routes
       app.use((req, res, next) => {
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -92,6 +92,15 @@ class MicrosoftGraphServer {
       const app = express();
       app.use(express.json());
       app.use(express.urlencoded({ extended: true }));
+      // Support TLS termination proxies
+      app.set('trust proxy', true);
+
+      const getBaseUrl = (req: Request): URL => {
+        const proto =
+          (req.headers['x-forwarded-proto'] as string)?.split(',')[0] || req.protocol;
+        const host = (req.headers['x-forwarded-host'] as string) || req.get('host');
```
Geod24, line 101: Use `get` instead of `headers` as it is case insensitive

Geod24 (commented):
Might want to add this variable in the [output at startup](https://github.com/Softeria/ms-365-mcp-server/blob/1e29081687eec28422a209dffe9b46cfa1ee17ff/src/server.ts#L74-L83) and more importantly in the [discovery configuration](https://github.com/Softeria/ms-365-mcp-server/blob/1e29081687eec28422a209dffe9b46cfa1ee17ff/src/server.ts#L121-L123).

Geod24:
@eirikb : FYI, we were trying to use this proxy in a production environment with open-webui. They added support for MCP as of v0.31.0 and I was hoping to expose our Entra tenant to it, but it would need a few fixes since IIUC your current HTTP use case is single user ? Our client (open-webui) itself has issue so we've been jumping between different servers to see what would work best, hence my interest in your project.

eirikb:
@Geod24 The point of the streamable HTTP version was multi-user, it should support OAuth (from agent) and it also supports direct tokens (if passed it just uses that token) - meaning it should work.
I haven't tested it much myself since my primary use is stdio, which is single user, but the HTTP mode should be multi user.

PR #133, opened 2025-10-09
feat: Add Bun binary compilation support

davenforce:
Add Bun Binary Compilation Support

  This PR adds the ability to compile the ms-365-mcp-server into standalone binaries that don't require Node.js
  installation.

  Changes

  Binary Compilation:
  - Added Bun build scripts for all platforms (macOS Intel/ARM, Linux x64/ARM64, Windows x64)
  - Auto-generate embedded endpoints data during build (bin/embed-endpoints.mjs)
  - Standalone binaries ~60MB each

  Code Changes for Binary Compatibility:
  - Made keytar optional with lazy loading and file-based fallback
  - Embedded endpoints.json as TypeScript constant (no runtime file reads)
  - Hardcoded version string (removed runtime package.json reads)
  - Fixed logging to use stderr (stdout reserved for MCP protocol)
  - Log files now write to ~/.ms365-mcp-server/logs/ (avoids permission issues)

  Documentation:
  - Updated README with binary installation option and build instructions
  - Added binary build scripts to package.json (build:binary:*)
  - Added prebuild hook to auto-generate embedded data

  Build Commands

  npm run build:binary:macos-arm    # macOS Apple Silicon
  npm run build:binary:macos         # macOS Intel  
  npm run build:binary:linux         # Linux x64
  npm run build:binary:windows       # Windows x64
  npm run build:binary:all           # All platforms

  Why Binary Distribution?

  - No Node.js installation required for end users
  - Simpler deployment (single executable file)
  - Smaller distribution size vs node_modules
  - Better for non-technical users

  Testing

  - Binary compiles successfully
  - Node.js version still works
  - Logging works correctly (stderr output)
  - Embedded endpoints load properly
  - Optional keytar fallback works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,6 @@ CLAUDE.md
 
 # MCP configuration (contains sensitive keys)
 .mcp.json
+
+# Generated endpoint data
+src/generated/endpoints-data.ts
--- a/README.md
+++ b/README.md
@@ -7,8 +7,22 @@ Microsoft 365 MCP Server
 A Model Context Protocol (MCP) server for interacting with Microsoft 365 and Microsoft Office services through the Graph
 API.
 
-## Prerequisites
+## Installation Options
 
+### Option 1: Standalone Binary (Recommended)
+
+Download pre-compiled binaries - no Node.js required:
+
+- **macOS (Apple Silicon)**: `ms365-mcp-server-macos-arm64`
+- **macOS (Intel)**: `ms365-mcp-server-macos`
+- **Linux (x64)**: `ms365-mcp-server-linux`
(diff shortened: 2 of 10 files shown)


eirikb:
Hi there.

I've been thinking about this PR a lot.
I honestly don't know what to do about it.
The server already works with bun, I don't know if adding support for binary matters that much, and to how many.
If it was a small change sure, but it's not.
And then I think about how much effort has gone into it, if you have done a ton of work to make it work then I would be hard pressed to not accept, but from the looks of it it might just be some prompting? I just don't know what to do.

eirikb:
Sorry, with no reply I'm going to close this now.

PR #135, opened 2025-10-18
fix: ensure retried requests go through full error handling and parsing

steffen-heil-secforge:
When a 401 error occurs and the token is refreshed, the retry request was bypassing all error handling (403, 404, etc.) and JSON parsing by returning early. This fix ensures the retried response flows through the same error checking and parsing logic as the initial request.

Changes:
- Changed `response` from const to let to allow reassignment
- Changed retry logic to reassign response instead of returning early

### DIFF
--- a/src/graph-client.ts
+++ b/src/graph-client.ts
@@ -54,7 +54,7 @@ class GraphClient {
     }
 
     try {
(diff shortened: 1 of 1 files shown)


eirikb:
Thanks :)

PR #136, opened 2025-10-18
feat: add excludeResponse parameter to reduce token usage

steffen-heil-secforge:
## Summary
Adds an optional `excludeResponse` parameter to all MCP tools that allows excluding the full response body and only returning success/failure indication.

## Use Case
This feature significantly reduces token usage in MCP responses when the caller only needs to know if an operation succeeded or failed, but doesn't need the actual response data.

**Example use cases:**
- Moving mail messages where you only care if the move succeeded
- Deleting items where confirmation is all that's needed
- Batch operations where individual response details aren't required

### Changes Made

1. **Tool Schema** (`src/graph-tools.ts`):
   - Added `excludeResponse` boolean parameter to all tool schemas (optional, default: false)
   - Added parameter filtering to exclude it from Graph API requests
   - Passes the option to GraphClient

2. **GraphClient** (`src/graph-client.ts`):
   - Added `excludeResponse` to `GraphRequestOptions` interface
   - Updated `formatJsonResponse()` to return minimal response when enabled
   - Added error handling to return minimal failure indication

### Response Format Examples

**Success with excludeResponse=true:**
```json
{
  "content": [{ "type": "text", "text": "{\"success\":true}" }],
  "isError": false
}
```

## Benefits
- Reduces token usage for operations where response data isn't needed
- Maintains backward compatibility (default behavior unchanged)
- Works for all operations (GET, POST, PUT, DELETE, PATCH)

## Note
The `isError` flag is **always** set correctly regardless of the `excludeResponse` setting, so clients can reliably detect failures while still benefiting from reduced response size.

### DIFF
--- a/src/graph-client.ts
+++ b/src/graph-client.ts
@@ -8,6 +8,7 @@ interface GraphRequestOptions {
   body?: string;
   rawResponse?: boolean;
   includeHeaders?: boolean;
+  excludeResponse?: boolean;
   accessToken?: string;
   refreshToken?: string;
 
@@ -164,7 +165,7 @@ class GraphClient {
       // Use new OAuth-aware request method
       const result = await this.makeRequest(endpoint, options);
 
-      return this.formatJsonResponse(result, options.rawResponse);
+      return this.formatJsonResponse(result, options.rawResponse, options.excludeResponse);
     } catch (error) {
       logger.error(`Error in Graph API request: ${error}`);
       return {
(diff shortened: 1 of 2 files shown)


eirikb:
Interesting, will this happen automatically when LLM decides it doesn't need the full result (just a OK, e.g., on item change)?

steffen-heil-secforge:
Github only copied the merge commit message, not the PR description. I copied it manually now. It contains more information.

My primary use of this is a mail sorting agent, that constantly flooded the context with the mails moved after classification. move-mail-message returned the whole mail, which seems to return the full message...

eirikb:
Yeah that was what I thought. I'm not sure the LLM will pick up on this automatically though, but if a user knows about then it can sprinkle the context

PR #191, opened 2026-01-22
fix: Excel range HTML entity encoding and skipEncoding support

aadorr:
### Issue
The `get-excel-range` tool consistently fails with `400 Bad Request - Error in query syntax` when attempting to read Excel cell ranges.

**Example failure:**
```
get-excel-range with address="A1:G10"
→ 400 Bad Request - Error in query syntax
```

**Root Cause:**
Microsoft Graph's Excel API uses function-style syntax for range operations:
```
/workbook/worksheets/{id}/range(address='A1:G10')
```

The `openapi-zod-client` code generator HTML-encodes special characters in path templates, converting:
- `=` → `&#x3D;`
- `'` → `&#x27;`  
- `(` → `&#x28;`
- `)` → `&#x29;`

This results in the generated client sending:
```
range(address=&#x27;A1:G10&#x27;)
```

Which Microsoft Graph API rejects as invalid syntax.

### Solution Applied

This PR implements a **post-processing approach** to fix the generated client code:

1. **Decode HTML entities** after generation (`generate-mcp-tools.mjs`)
   - Converts `&#x27;` back to `'`, `&#x3D;` back to `=`, etc.

2. **Convert to template literals** for TypeScript compatibility
   - Changes `path: '/...range(address=':address')...'` 
   - To `path: `/...range(address=':address')...``
   - This avoids nested single-quote syntax errors

3. **Add `skipEncoding` parameter support** (`graph-tools.ts`)
   - Prevents URL-encoding of the `address` parameter value
   - `A1:G10` stays as-is instead of becoming `A1%3AG10`

4. **Configure `skipEncoding` for Excel range endpoint** (`endpoints.json`)
   - Marks the `address` parameter to skip encoding

### Testing
Verified with Postman that the corrected URL works:
```
/range(address='A1:G10')  ✅ Returns Excel data
/range(address=&#x27;A1:G10&#x27;)  ❌ 400 Bad Request
```

### Alternative Solution
**Note:** This issue could also be resolved upstream in `openapi-zod-client` itself by:
- Adding an option to disable HTML entity encoding for path templates, OR
- Detecting function-style API patterns and preserving literal characters

However, that would require changes to the code generator library. This PR provides an immediate fix by post-processing the generated output.

### Impact
- ✅ Fixes `get-excel-range` tool (previously completely broken)
- ✅ Generic solution works for any future function-style Graph API endpoints
- ✅ No breaking changes to existing functionality

### Files Changed
- `bin/modules/generate-mcp-tools.mjs` - HTML entity decoding post-processing
- `src/graph-tools.ts` - Add `skipEncoding` parameter support  
- `src/endpoints.json` - Configure `skipEncoding` for `address` parameter
- `src/graph-client.ts` - Minor formatting (empty line for readability)

### DIFF
--- a/bin/modules/generate-mcp-tools.mjs
+++ b/bin/modules/generate-mcp-tools.mjs
@@ -40,11 +40,11 @@ export function generateMcpTools(openApiSpec, outputDir) {
     console.log('Decoding HTML entities in path patterns...');
     // openapi-zod-client HTML-encodes special characters in path patterns
     // This breaks Microsoft Graph function-style APIs like range(address='A1:G10')
-    clientCode = clientCode.replace(/&#x3D;/g, '=');  // Decode = sign
-    clientCode = clientCode.replace(/&#x27;/g, "'");  // Decode single quote
-    clientCode = clientCode.replace(/&#x28;/g, '(');  // Decode left paren
-    clientCode = clientCode.replace(/&#x29;/g, ')');  // Decode right paren
-    clientCode = clientCode.replace(/&#x3A;/g, ':');  // Decode colon
+    clientCode = clientCode.replace(/&#x3D;/g, '='); // Decode = sign
+    clientCode = clientCode.replace(/&#x27;/g, "'"); // Decode single quote
+    clientCode = clientCode.replace(/&#x28;/g, '('); // Decode left paren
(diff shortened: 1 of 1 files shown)


eirikb:
Thanks, LGTM. Just fix the formatting and we can merge this :)
The additional logging stands out, but I see we already do this further down so all good.

PR #210, opened 2026-01-30
fix: resolve stdio handshake failure (#204) and admin consent token overwrite (#209)

### LINKED ISSUE #204: mcporter fails to connect to @softeria/ms-365-mcp-server over stdio
Hi, I’m trying to use @softeria/ms-365-mcp-server together with mcporter (via the Clawdbot m365 skill), and I’m consistently getting a stdio handshake failure even though the server and Microsoft 365 auth both work fine.
Environment

    OS: Ubuntu (desktop, 64‑bit)
    @softeria/ms-365-mcp-server: current npm release (npx -y @softeria/ms-365-mcp-server)

mcporter: 0.7.3 CLI

Clawdbot: using the m365 skill from cvsloane/m365-skill (Microsoft 365 skill for Clawdbot)

    
What works

    Microsoft 365 authentication and Graph access via the bundled CLI:

bash
cd ~/clawd/skills/ms365

python3 ms365_cli.py login
python3 ms365_cli.py status
python3 ms365_cli.py user
python3 ms365_cli.py accounts

All of those return success JSON and my user/account details.
​

    Real data from Microsoft 365:

Commands return actual email and calendar events, so Graph permissions and tokens are valid.
​

​

    Running the MCP server directly:

bash
npx -y @softeria/ms-365-mcp-server

This starts the server and it stays running; it does not immediately exit or crash.
​
​
mcporter configuration

My mcporter entry (simplified) points to the npm package via stdio:

text
{
  "servers": {
    "m365": {
      "transport": "stdio",
      "command": "npx",
      "args": ["-y", "@softeria/ms-365-mcp-server"]
    }
  }
}

Running:

bash
mcporter list

shows:

    m365 (offline – unable to reach server, 2.8s)

    Listed 1 server (0 healthy; 1 offline).

​
If I try a tool call (e.g. mcporter call m365.someTool from Clawdbot/mcporter), I get an MCP error similar to:

    MCP error -32000: Connection closed

Even with the per‑server timeout set to 30 seconds, mcporter gives up after about 2.8 seconds, so it looks like an early handshake failure rather than a simple timeout.
What I’ve tried

    Verified Node version and that npx -y @softeria/ms-365-mcp-server runs cleanly on its own.

Confirmed Microsoft 365 auth and data access via ms365_cli.py (so CLIENT_ID / SECRET / TENANT_ID etc. are correct, and Graph is reachable).

Increased mcporter per‑server timeout to 30 seconds; mcporter still reports offline – unable to reach server, 2.8s.

    ​

Suspected issue

Given that:

    The server runs fine when started directly.

    Microsoft 365 auth and Graph calls work via the provided CLI.

    Only mcporter’s stdio connection reports offline / Connection closed within a couple of seconds.

It seems like there is an incompatibility or bug in the stdio MCP handshake between mcporter and @softeria/ms-365-mcp-server (or how the Clawdbot m365 skill wires them together).
Questions

    Is there a known issue or required flag for running @softeria/ms-365-mcp-server under mcporter via stdio (e.g. needing --http or other options)?

Do you have an example mcporter config that is known to work with this server?

Is there extra logging I can enable on the server or on mcporter to capture the exact handshake failure?

### LINKED ISSUE #209: Admin consent flow overwrites user token - blocks multi-user scenarios
## Problem

When a non-admin user attempts to authenticate, the OAuth flow requires admin consent. When an admin grants consent in the same browser session, the admin's token overwrites the original user's token, making it impossible for non-admin users to authenticate.

## Steps to Reproduce

1. Run `npx @softeria/ms-365-mcp-server --login`
2. Sign in as a non-admin user (e.g., `sallie@example.com`)
3. Microsoft prompts "Need admin approval"
4. Click "Have an admin account? Sign in with that account"
5. Admin signs in and grants consent
6. Result: Token is now for the admin account, not the original user

## Expected Behavior

- Non-admin users should be able to authenticate after admin consent is granted
- Admin consent should be a separate flow that doesn't replace the user token
- Or: Use user-delegated permissions that don't require admin consent

## Current Workarounds Attempted

- Pre-granting admin consent via `https://login.microsoftonline.com/{tenant}/adminconsent?client_id=084a3e9f-a9f4-43f7-89f9-d229cf97853e` - still prompts for admin approval
- Clicking "Return to the application without granting consent" - results in error page

## Environment

- `@softeria/ms-365-mcp-server@0.32.0`
- Ubuntu 24.04 (Azure VM)
- Microsoft Edge browser

## Suggestions

1. Support user-delegated permissions only (no admin consent required) for personal mailbox/calendar/tasks access
2. Separate admin consent flow from user authentication
3. Allow pre-granted tenant consent to be recognized
4. Better error handling when consent is declined

Thank you for this useful MCP server!

BenGWeeks:
## Summary

This PR fixes two critical issues affecting MoltBot/ClawBot compatibility:

### Issue #204: stdio Handshake Failure
- **Root cause**: Logger output to stdout/stderr after stdio transport connection corrupted the MCP JSON-RPC handshake
- **Fix**: 
  - Console logging now only enabled in HTTP mode by default (or with `-v` flag)
  - All Winston log levels route to stderr via `stderrLevels` configuration
  - Removed the `logger.info('Server connected to stdio transport')` call that occurred after connection

### Issue #209: Admin Consent Token Overwrite  
- **Root cause**: When admin consent is required, the admin's token could replace the user's token since both accounts get cached and `getCurrentAccount()` fell back to `accounts[0]`
- **Fix**:
  - Track the currently selected account before device code authentication
  - Detect when a different account is returned after authentication
  - Preserve the original account selection when account mismatch occurs
  - Warn users with clear guidance via logs and MCP tool output

## Files Changed
- `src/logger.ts` - Add `stderrLevels` to route all logs to stderr
- `src/server.ts` - Enable console logging only in HTTP mode or with `-v`; remove post-connection log
- `src/auth.ts` - Add account validation after device code flow

## Test plan
- [x] All existing tests pass (`npm run test`)
- [x] Lint passes (`npm run lint`)
- [x] Build succeeds (`npm run build`)
- [x] Format check passes (`npm run format:check`)
- [ ] Test stdio mode: Run `npx tsx src/index.ts` and verify no output to stdout before MCP handshake
- [ ] Test with MoltBot/ClawBot: Confirm MCP connection establishes properly
- [ ] Test admin consent flow: Login as non-admin, trigger admin consent, verify warning message

Closes #204
Closes #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,40 @@ To access work/school features (Teams, SharePoint, etc.), enable organization mo
 Organization mode must be enabled from the start to access work account features. Without this flag, only personal
 account features (email, calendar, OneDrive, etc.) are available.
 
+## User-Only Mode (No Admin Consent)
+
+If you're using a work/school account but don't have tenant admin approval, use `--user-only` to request only user-delegated scopes that don't require admin consent:
+
+```bash
+npx @softeria/ms-365-mcp-server --login --user-only
+```
+
+Or in Claude Desktop configuration:
+
+```json
+{
+  "mcpServers": {
+    "ms365": {
+      "command": "npx",
+      "args": ["-y", "@softeria/ms-365-mcp-server", "--user-only"]
+    }
+  }
+}
+```
+
+This filters out scopes ending with `.All` or `.Shared` (like `Files.Read.All`, `User.Read.All`) which require admin consent.
+
+**User-only scopes that work without admin consent:**
+
+- `Mail.Read`, `Mail.ReadWrite`, `Mail.Send`
+- `Tasks.ReadWrite`
+- `Chat.Read`, `ChatMessage.Read`, `ChatMessage.Send`
+- `Calendars.Read`, `Calendars.ReadWrite`
+- `Files.Read`, `Files.ReadWrite`
+- `User.Read`
+
+> **Note**: Some features (Teams channels, SharePoint sites, user directory) won't be available in user-only mode since they require admin-consent scopes.
+
 ## Shared Mailbox Access
 
 To access shared mailboxes, you need:
@@ -456,6 +490,7 @@ The following options can be used when running ms-365-mcp-server directly from t
 --work-mode       Alias for --org-mode
 --force-work-scopes Backwards compatibility alias for --org-mode (deprecated)
 --cloud <type>    Microsoft cloud environment: global (default) or china (21Vianet)
+--user-only       Request only user-delegated scopes (no admin consent required)
 ```
 
 ### Server Options
@@ -482,6 +517,7 @@ Environment variables:
 - `ENABLED_TOOLS`: Filter tools using a regex pattern (alternative to --enabled-tools flag)
 - `MS365_MCP_ORG_MODE=true|1`: Enable organization/work mode (alternative to --org-mode flag)
 - `MS365_MCP_FORCE_WORK_SCOPES=true|1`: Backwards compatibility for MS365_MCP_ORG_MODE
+- `MS365_MCP_USER_ONLY=true|1`: Request only user-delegated scopes (alternative to --user-only flag)
 - `MS365_MCP_OUTPUT_FORMAT=toon`: Enable TOON output format (alternative to --toon flag)
 - `MS365_MCP_CLOUD_TYPE=global|china`: Microsoft cloud environment (alternative to --cloud flag)
 - `LOG_LEVEL`: Set logging level (default: 'info')


BenGWeeks:
## Testing Results from Deployment

I deployed this PR to test on a real MoltBot/ClawdBot setup. Here are the findings:

### What Works
- The stdio fix (#204) - console logging is properly routed to stderr
- The token overwrite detection code (#209) is present in the build

### What Doesn't Work
The core issue is that **the app requests admin-consent-required scopes** even for basic personal use cases:

- `Channel.ReadBasic.All`, `ChannelMessage.Read.All`
- `Files.Read.All`, `Sites.Read.All`  
- `Team.ReadBasic.All`, `TeamMember.Read.All`, `User.Read.All`
- `*.Shared` variants (shared mailboxes/calendars)

**Test flow:**
1. Run `ms-365-mcp-server --login` as non-admin user
2. Microsoft prompts "Need admin approval"
3. Click "Return to the application without granting consent"
4. Result: **Fails with "This is not the right page" error**

The token overwrite protection only kicks in AFTER successful authentication. But auth fails before that because the scopes require admin consent.

### Suggested Enhancement
Add a `--user-only` or `--no-admin-scopes` flag that filters out admin-consent-required scopes:

```typescript
const ADMIN_CONSENT_SCOPES = /\.(All|Shared)$/;
if (options.userOnly) {
  scopes = scopes.filter(s => !ADMIN_CONSENT_SCOPES.test(s));
}
```

**User-only scopes that work without admin consent:**
- `Mail.Read`, `Mail.ReadWrite`, `Mail.Send`
- `Tasks.ReadWrite`
- `Chat.Read`, `ChatMessage.Read`, `ChatMessage.Send`
- `Calendars.Read`, `Calendars.ReadWrite`
- `Files.Read`, `Files.ReadWrite`
- `User.Read`

This would enable personal/individual use without requiring tenant admin approval - a common use case for AI agents like MoltBot/ClawdBot.

BenGWeeks:
I spent my evening going round in login / authentication loops!!!

I actually think this MCP would benefit from being split up into parts as the context it loads is huge ... e.g. mcp-microsoft-teams, mcp-microsoft-outlook, mcp-microsoft-todo, etc. This would be much more efficient with tokens and less prone to agent mis-use.

## Testing Report

Tested this PR extensively with a dedicated Azure VM running an AI agent (Sallie) that needs MS365 access.

### What works well
- The `--user-only` flag correctly filters out admin-consent-required scopes at runtime
- The admin consent detection logic in `acquireTokenByDeviceCode` works when a token is already cached

### Issues discovered

**1. Softeria's app registration still requires admin consent**

Even with `--user-only`, Azure AD still prompts for admin consent because the **app registration** declares admin-consent scopes (like `User.Read.All`, `Files.Read.All`). The `--user-only` flag filters what we *request*, but Azure AD checks what the *app* declares.

**Workaround:** Users need to create their own Azure app registration with only user-delegated scopes. This is the only way to avoid the admin consent requirement entirely.

**2. Missing `offline_access` scope causes refresh token failures**

After successful login, subsequent `--verify-login` calls fail with "Silent token acquisition failed" because no refresh token was issued. The fix is to always include `offline_access` in the scope request:

```typescript
// In buildScopesFromEndpoints or similar
if (!scopes.includes('offline_access')) {
  scopes.push('offline_access');
}
```

**3. Admin consent stores the wrong user's token**

When an admin grants consent, their token gets cached instead of the intended user's. The current detection only works if a token is already cached. 

**Suggestion:** Add a `--login-hint <email>` flag that validates the authenticated user matches the expected user, rejecting the token if there's a mismatch:

```typescript
if (expectedUser && response?.account?.username?.toLowerCase() !== expectedUser.toLowerCase()) {
  throw new Error(`Expected user "${expectedUser}" but "${response.account.username}" authenticated`);
}
```

### Recommendation

Consider documenting that users who want to avoid admin consent entirely should create their own Azure app registration with only user-delegated scopes, and set `MS365_MCP_CLIENT_ID` / `MS365_MCP_TENANT_ID` environment variables.

FlyGoat:
For the `--user-only` flag, it seems like different organisations has different preferences on what would require admin consent. For example my University is pretty restrictive and only filtering out `.All` is not sufficient.

My suggestion would be having a parameter to allow us setting up filters for scopes.

PR #214, opened 2026-02-01
feat: Add batch email operations for efficient bulk processing

NightOwlCoder:
## Summary

Adds 4 new batch tools that use Microsoft Graph's `$batch` endpoint to process up to 20 emails per API call (20x more efficient than individual calls).

## New Tools

| Tool | Description |
|------|-------------|
| `batch-delete-messages` | Delete up to 20 messages per call |
| `batch-move-messages` | Move up to 20 messages to a folder |
| `batch-archive-messages` | Archive up to 20 messages |
| `batch-mark-read` | Mark up to 20 messages as read/unread |

## Use Case

Processing large inboxes (6,000+ emails) with current single-message tools is slow and expensive. Batch operations reduce API calls by 95%.

## Implementation

- New `src/batch-tools.ts` with all 4 tools
- Uses existing `GraphClient.graphRequest()` to call `/$batch` endpoint
- Registered in `server.ts` (disabled in read-only mode)
- Requires `Mail.ReadWrite` scope (same as existing mail tools)

## Testing

Tested with personal Outlook.com account:
- ✅ `batch-mark-read`: Marked messages as read/unread
- ✅ `batch-archive-messages`: Moved messages to Archive folder
- ✅ `batch-move-messages`: Moved messages between folders  
- ✅ `batch-delete-messages`: Deleted messages

## Graph Batch Format

```json
POST /$batch
{
  "requests": [
    { "id": "1", "method": "DELETE", "url": "/me/messages/{id1}" },
    { "id": "2", "method": "DELETE", "url": "/me/messages/{id2}" }
  ]
}
```

Response includes per-message status for error handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ MS365_MCP_OUTPUT_FORMAT=toon npx @softeria/ms-365-mcp-server
 
 ## Supported Services & Tools
 
+**Email Batch Operations** _(new)_  
+<sub>batch-delete-messages, batch-move-messages, batch-archive-messages, batch-mark-read</sub>  
+Efficiently process up to 20 messages per API call using Microsoft Graph's $batch endpoint.
+
 ### Personal Account Tools (Available by default)
 
 **Email (Outlook)**  
--- a/src/batch-tools.ts
+++ b/src/batch-tools.ts
@@ -1,0 +1,326 @@
+import { z } from 'zod';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import GraphClient from './graph-client.js';
+import logger from './logger.js';
+
+const MAX_BATCH_SIZE = 20;
+
+interface BatchRequest {
+  id: string;
+  method: string;
+  url: string;
+  headers?: Record<string, string>;
+  body?: unknown;
+}
+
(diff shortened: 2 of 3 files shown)


eirikb:
Hi there, and thanks! :slightly_smiling_face: 

I've gone through this PR now, and I think we need to discuss it a bit.  
Adding batch support is a great idea, and we should definitely do this.  
Today all endpoints are mapped up in **endpoints.json** file, and I think the endpoints you have manually added should be there as well. However, batch is kind of supported on all graph endpoints right, so we need a new pattern.  
How about: 

```json
  {
    "pathPattern": "/me/messages/{message-id}",
    "method": "delete",
    "toolName": "delete-mail-message",
    "scopes": ["Mail.ReadWrite"]
  },
  {
    "pathPattern": "/me/messages/{message-id}",
    "method": "batch:delete",
    "toolName": "batch-delete-mail-messages",
    "scopes": ["Mail.ReadWrite"]
  }
```

Why I think this is important is because a lot the other logic surrounds the endpoints.json file, such as filtering and discovery.  
With this one should be able to disable them by `--enabled-tools "^(?!batch)" ` (or enable if we flip that).

What do you think?

PR #216, opened 2026-02-02
feat: Add bodyMaxLength parameter to truncate email bodies

NightOwlCoder:
## Problem

When listing emails via `list-mail-messages`, the full HTML body is returned for each email. Newsletter emails can be 50KB+ each, so:

- 50 emails x 50KB = 2.5MB per API call
- LLM context fills up in ~15 messages
- Email triage workflows become impossible

## Solution

Add `bodyMaxLength` parameter to truncate email bodies at the MCP server level.

### Changes

1. **New parameter** for `list-mail-messages` and `list-mail-folder-messages`:
   - `bodyMaxLength` (number, optional) - Maximum characters for `body.content`
   - Recommended: 1024 for triage workflows

2. **Truncation with marker** - appends original size for reference

3. **Logging** - Server logs truncation count for debugging

### Usage

```json
{"top": 50, "bodyMaxLength": 1024}
```

### Results

- 50 emails: 2.5MB to ~100KB (96% reduction)
- Context stays manageable for LLM workflows
- Original length preserved in marker for reference

### Implementation

- Parameter-driven: No hardcoded endpoint detection. If param is passed, truncation runs.
- Safe: Only truncates if body.content exists and exceeds limit
- Non-breaking: Parameter is optional, full body returned if omitted

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ MS365_MCP_OUTPUT_FORMAT=toon npx @softeria/ms-365-mcp-server
 
 ## Supported Services & Tools
 
+**Email Batch Operations** _(new)_  
+<sub>batch-delete-messages, batch-move-messages, batch-archive-messages, batch-mark-read</sub>  
+Efficiently process up to 20 messages per API call using Microsoft Graph's $batch endpoint.
+
 ### Personal Account Tools (Available by default)
 
 **Email (Outlook)**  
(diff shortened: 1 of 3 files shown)


eirikb:
Hi. This is a good idea, but could we do it differently? I think there is already a field called `bodyPreview`, we only need to make the LLM understand it can be wise to use this - perhaps the fix is a simple as noting that in `llmTip`?

PR #240, opened 2026-02-18
feat: multi-account support via per-tool-call account parameter

marlian:
## Summary
Adds multi-account support: a single server instance can serve N Microsoft accounts via an optional `account` parameter injected into all tool schemas when more than one MSAL account is cached.

## Changes
- **`src/auth.ts`**: `isMultiAccount()` detects >1 MSAL cached account; `getTokenForAccount(identifier?)` resolves by email (case-insensitive) or `homeAccountId`; `isOAuthModeEnabled()` accessor for OAuth mode detection
- **`src/graph-tools.ts`**: `account` param injection into tool schemas when multi-account mode is active; per-tool-call token resolution in `executeGraphTool`; token redaction in log output
- **`src/server.ts`**: `multiAccount` detection at `initialize()`; pass-through to tool registration
- **`README.md`**: Multi-account usage documentation with examples

## Motivation
MCP multiplexers (like Legate) that manage multiple Microsoft accounts currently need to spawn N server instances — one per account. With 2 accounts × 109 tools = 218 tools in the LLM context, which causes model confusion and resource waste. This change enables a single instance to serve all accounts via the `account` parameter, reducing tool count to 109.

## Design
- Reuses MSAL's existing multi-account token cache (login multiple accounts via repeated `--login`)
- `getTokenForAccount()` resolution: username (email, case-insensitive) → `homeAccountId` → auto-select (single account) → `selectedAccountId` (multi-account default) → helpful error
- Token passed via existing `options.accessToken` override in `GraphClient.makeRequest()` — **no graph-client changes needed**
- OAuth/HTTP mode: account resolution skipped, request context drives token selection
- `account` parameter stripped from tool params before Graph API processing

## Backward compatibility
- **Single-account setups**: zero change — no `account` param in schemas, auto-selects
- **OAuth/HTTP mode**: unchanged — request context drives token selection
- **CLI**: existing `--login`, `--list-accounts`, `--select-account` work as before

## Security
- Access tokens redacted from log output (prevents credential leakage via `JSON.stringify(options)`)
- Multi-account no-identifier path does NOT fall back to `accounts[0]` — requires explicit `selectedAccountId` or throws
- OAuth mode correctly preserved — per-request tokens not overridden

## Test status
- Build: ✅
- 28/28 tests pass ✅
- 3 test suites fail pre-existingly (missing `generated/client.js`)


marlian:
> What if, on multiple accounts, there would be a tool for listing accounts? I bet LLMs would handle that smartly

That's a great idea! If you think it's worth it, I'd happily add it. That would give us three layers of account discovery, progressively richer:

| Layer | Mechanism | How it works | Status |

1 ) Self-healing error | LLM omits account → error response lists available accounts → LLM retries with correct value | Already in this PR |
2 ) Dynamic enum on schema | account field populated with cached usernames at startup → LLM sees options directly in the tool definition, zero round-trips | Can add |
3 ) list-accounts tool | Dedicated MCP tool the LLM can call proactively to discover accounts before making any call | Can add |

Each layer covers a different failure mode: layer 1 is the safety net, layer 2 gives the LLM the info upfront, and layer 3 lets it ask explicitly when it needs to.

Want me to implement layers 2 and 3 as well?

eirikb:

PR #259, opened 2026-03-04
feat: 111 llmTips, OData describe overrides, Windows chmod fix, tests, README Graph API tips

RaoulJacobs:
## Summary

- **111 llmTips** added across all tool categories in \`endpoints.json\` — helps LLMs pick the right tool and use correct parameters
- **OData parameter describe() overrides** in \`graph-tools.ts\` — \`\$filter\`, \`\$search\`, \`\$select\`, \`\$orderby\`, \`\$top\`, \`\$skip\`, \`\$count\` now have human-readable descriptions visible in MCP tool definitions
- **Windows compatibility fix** in \`tsup.config.ts\` — \`chmod +x\` is skipped on Windows to prevent build errors
- **Kebab-case path param fix** in \`graph-tools.ts\` — \`message-id\` (kebab) now correctly resolves to \`messageId\` (camelCase) in URL path substitution. Fixes \`Id is malformed\` / \`Empty Payload\` errors for all endpoints with path params (get/update/delete message, attachments, etc.)
- **9 new tests** in \`src/__tests__/graph-tools.test.ts\` — OData parameter parsing, pagination, path param normalization (kebab→camelCase)
- **README Graph API Tips & Known Quirks** section — OData filtering syntax, pagination patterns, search limitations
- **Restore \`get-root-folder\`** as deprecated (non-breaking) with llmTip pointing to \`list-children\`
- **update-mail-message llmTip** updated to document correct path param name and body usage

## Test plan

- [ ] \`npm run build\` passes on Linux/Mac and Windows
- [ ] \`npm test\` — 9 new tests pass
- [ ] MCP tool definitions show OData parameter descriptions
- [ ] llmTips visible in MCP tool listings
- [ ] \`message-id\` and \`messageId\` both work as path param for message endpoints

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -4,28 +4,28 @@
     "method": "get",
     "toolName": "list-mail-messages",
     "scopes": ["Mail.Read"],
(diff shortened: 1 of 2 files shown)


eirikb:
Hi. This is quite the PR.

The reason we haven't added llmTips all over, or tried to be restrictive, is because we believe it eats context on startup.

Do you have some numbers on how many tokens this change adds to the context on startup? (not limiting tools, not using `--discovery` mode).

eirikb:
Ok I've been thinking.

There are now two reasons for the `llmTip`s:
1. Tell the LLM about bad spec (_`$select` on todo doesn't work_)
2. Guide the LLM (_wrap `$search` in quotes_)

In theory we shouldn't need `llmTip`s, everything should work perfectly based on the spec - which also has descriptions.

The idea behind `llmTip` was a quick and dirty way to inform LLM about errors (or missing parts) in the specification. The only problem is that LLMs see this field and throw a lot of stuff into it because they love generating text and here is a possibility to do just that. Can also be the LLM doesn't grasp fully that tools are fully explained through spec, and thinks this field should be used to do basically all the work.

What do you think? Do all the `llmTip`s in this PR expand the LLM knowledge beyond what is available in spec (and what should be part of the LLM)?  
Did your agent do any research per tool, or are all the tips from the LLM itself, if so then other LLMs should be able to do the same automatically.

For example `$select`. It is littered over the `llmTip`s, and perhaps we have to hint to the LLM every time that it should use `$select`, which is sad, but even then the available fields should be in the spec (if it's not then perhaps we should find out why).

PR #269, opened 2026-03-11
fix: OData param encoding, path param injection, ImmutableId header

RaoulJacobs:
## Summary

A set of bug fixes for Graph API parameter handling discovered while using the MCP server in production:

- **Kebab-case path params**: `message-id` passed by LLMs was silently ignored → path stayed `:messageId` → `Id is malformed` or `Empty Payload` errors. Now normalizes kebab-case to camelCase before lookup and substitution.
- **Auto-inject missing path params**: Tools whose path params are not declared in `tool.parameters` (generated client omits them) now auto-inject them from the path pattern, preventing 404s.
- **`$` in query param keys not encoded**: `$select` was being encoded to `%24select` → Graph API 400. Now preserves `$` prefix.
- **Commas in OData values encoded**: Literal commas in filter values (e.g. `$select=id,displayName`) were encoded as `%2C` → Graph API error. Now preserves literal commas.
- **`Prefer: IdType=ImmutableId` header**: Base64-encoded IDs containing `/` caused `RequestBroker--ParseUri` 400 errors when used in path segments. Adding this header tells Graph to return/accept immutable IDs (base64url, no `/` or `+`).

## Test plan

- [x] Unit tests added for kebab-case normalization and path param auto-injection
- [ ] Manual test: call `get-mail-message` with `message-id` param (kebab) — should resolve correctly
- [ ] Manual test: call `list-mail-messages` with `$select=id,subject` — should not encode commas
- [ ] Manual test: call any todo endpoint — should not get RequestBroker errors

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -468,14 +468,14 @@
     "method": "get",
     "toolName": "list-todo-task-lists",
     "scopes": ["Tasks.Read"],
-    "llmTip": "Lists all To Do task lists. Use $select=id,displayName to reduce response size. Returns todoTaskList-id needed for all task operations. The default list is typically called 'Tasks'."
+    "llmTip": "Lists all To Do task lists. IMPORTANT: $select is NOT supported on this endpoint (Microsoft limitation) — do not pass select parameter. Returns todoTaskList-id needed for all task operations. The default list is typically called 'Tasks'."
   },
   {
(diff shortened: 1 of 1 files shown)


RaoulJacobs:
Rebased on latest upstream. Conflict resolved. Ready to review.

Issue #287, opened 2026-03-20
Need to reregister OWUI client after X amount of time


HenkieTenkie62:
I'm connected with http mode, registered OAuth 2.1, --enable-dynamic-registration and --org-mode and use the tool through Open WebUI.
Normally the login to Microsoft times out/expires and you need to login again.
Most of the time this is handled perfectly fine by the tool and you are sent to a Microsoft login page when enabling the tool in OWUI interface.
But after an X amount of time instead of showing the Microsoft login it shows an empty page with an error.

To fix this I need to reregister client in OWUI:
<img width="526" height="543" alt="Image" src="https://github.com/user-attachments/assets/d9edd24d-38f5-4ac0-a8d6-9927df2fe578" />

Is this something the tool should handle?
I just fixed it for the Nth time and forgot to look at logs and make a screenshot.
I think it occurs weekly (after reboot?), so maybe I can capture some more info next week.

HenkieTenkie62:
Same again after updating OWUI this morning to v0.9.1;
Maybe it is only because of upgrading Open WebUI.

Logs from OWUI:

```logs
2026-04-22 11:17:36.679 | WARNING  | open_webui.utils.oauth:get_oauth_token:767 - No OAuth session found for user <userkey/id>, client_id mcp:ms-365

2026-04-22 11:17:36.681 | WARNING  | open_webui.utils.oauth:get_oauth_token:767 - No OAuth session found for user <userkey/id>, client_id mcp:ms-365-<other instance>
```

OWUI had some OAuth related upgrades in v0.9.0 and ms-365-mcp as well so I updated both.
My ms-365-mcp instance is now on latest (v0.85.0).

Is it expected behaviour to need to reregister the tool after client update?

eirikb:
This should have been fixed in #406 . Feel free to re-open if still doesn't work.

The errors in your v0.85.0 logs - Error decrypting data and No OAuth session found for user … client_id mcp:ms-365 - are OWUI own encrypted OAuth store losing its key across upgrades/reboots, not something this server can fix.

Re-registering after an OWUI update isn't expected, but it's an OWUI state issue.

PR #337, opened 2026-03-30
fix: return new refresh token to client after server-side 401 retry

### LINKED ISSUE #336: Bug: Server-side 401 retry silently consumes refresh token without returning new one to client (token rotation breakage)
### Summary

In HTTP mode, when a Microsoft Graph API call returns `401`, the server automatically refreshes the access token using the client-provided refresh token and retries the request. However, **the new refresh token issued by Microsoft during this process is never returned to the client**. Since Microsoft uses [refresh token rotation](https://learn.microsoft.com/en-us/entra/identity-platform/refresh-tokens#refresh-token-lifetime), the client's original refresh token becomes invalid, effectively breaking the client's session.

### Root Cause

**File:** `src/graph-client.ts` (lines 65–72)

```typescript
if (response.status === 401 && refreshToken) {
  // Token expired, try to refresh
  const newTokens = await this.refreshAccessToken(refreshToken);
  accessToken = newTokens.accessToken;

  // Retry the request with new token
  response = await this.performRequest(endpoint, accessToken, options);
}
```

The `refreshAccessToken()` method (lines 125–151) correctly retrieves both the new access token **and** the new refresh token from Microsoft:

```typescript
return {
  accessToken: response.access_token,
  refreshToken: response.refresh_token,  // ← new RT obtained but never used
};
```

However, only `newTokens.accessToken` is used for the retry. `newTokens.refreshToken` is **silently discarded** — there is no mechanism (response header, response body, etc.) to communicate it back to the client.

### Sequence Diagram

```
Client                        MCP Server                     Microsoft
  |                               |                              |
  |-- POST /mcp ----------------->|                              |
  |   Authorization: Bearer AT_old|                              |
  |   x-microsoft-refresh-token:  |                              |
  |                       RT_v1   |                              |
  |                               |-- Graph API (AT_old) ------->|
  |                               |<-- 401 Unauthorized ---------|
  |                               |                              |
  |                               |-- POST /token (RT_v1) ------>|
  |                               |<-- AT_new + RT_v2 -----------|
  |                               |   (RT_v1 is now INVALIDATED) |
  |                               |                              |
  |                               |-- Graph API (AT_new) ------->|
  |                               |<-- 200 OK -------------------|
  |                               |                              |
  |<-- MCP response (data only) --|                              |
  |    ⚠ RT_v2 is NOT returned    |                              |
  |                               |                              |
  | [Client still holds RT_v1 — now invalid]                     |
  |                               |                              |
  |-- Next request with RT_v1 --->|                              |
  |                               |-- POST /token (RT_v1) ------>|
  |                               |<-- 400 Invalid Grant --------|
  |                               |                              |
  |<-- ❌ Authentication failure --|                              |
```

### Impact

- Client's refresh token becomes **silently invalid** after any server-side 401 retry
- No error is surfaced — the current request succeeds, but the **next** request fails
- Client must perform full re-authentication (OAuth consent flow)
- Particularly problematic for long-running sessions where access tokens expire naturally (~1 hour)

### Possibly Related Issues

- **#287** — "Need to reregister OWUI client after X amount of time" — This may be a symptom of this exact issue. After a server-side token refresh occurs, the OWUI client's stored refresh token becomes invalid, eventually forcing re-registration.
- **#298** (closed) — Token cache desync between keytar and file storage — Different root cause but similar symptom (unexpected re-authentication).

### Suggested Fix

Return the new refresh token to the client via a response header after server-side refresh:

**Option A: Response header (minimal change)**

In `graph-client.ts`, after the 401 retry succeeds, set a response header:

```typescript
if (response.status === 401 && refreshToken) {
  const newTokens = await this.refreshAccessToken(refreshToken);
  accessToken = newTokens.accessToken;

  // Store new refresh token in request context so it can be returned to client
  if (newTokens.refreshToken) {
    updateRequestTokens({ refreshToken: newTokens.refreshToken });
  }

  response = await this.performRequest(endpoint, accessToken, options);
}
```

Then in the HTTP handler (`server.ts`), read the updated context and set a response header:

```typescript
res.setHeader('X-Microsoft-New-Refresh-Token', updatedRefreshToken);
```

**Option B: Don't retry on server side (safest)**

Remove the server-side 401 retry entirely and let the client handle token refresh. This keeps the server truly stateless and avoids the token rotation problem altogether. Return a 401 with a clear error so the client knows to refresh and retry.

### Environment

- **Affects:** All versions with HTTP mode
- **Mode:** HTTP mode only (stdio mode manages tokens differently)
- **Relevant files:**
  - `src/graph-client.ts` — 401 retry logic (lines 65–72), `refreshAccessToken()` (lines 125–151)
  - `src/lib/microsoft-auth.ts` — `refreshAccessToken()` function (lines 94–133)
  - `src/server.ts` — HTTP handlers (lines 384–470)
  - `src/request-context.ts` — AsyncLocalStorage context (lines 3–12)

badmintoncryer:
## Summary

Fixes #336

When the server refreshes a token due to a `401` response from Microsoft Graph API in HTTP mode, the new refresh token was **silently discarded**. Due to Microsoft's [refresh token rotation](https://learn.microsoft.com/en-us/entra/identity-platform/refresh-tokens#refresh-token-lifetime), this invalidated the client's stored refresh token, breaking subsequent requests.

### `src/request-context.ts`
- Added `onTokenRefreshed` optional callback to `RequestContext` interface
- Added `notifyTokenRefreshed()` helper function that invokes the callback within the current async context

### `src/graph-client.ts`
- After a successful token refresh on 401 retry, calls `notifyTokenRefreshed()` with the new refresh token (+4 lines)

### `src/server.ts`
- Both GET and POST `/mcp` handlers now pass an `onTokenRefreshed` callback to `requestContext.run()` that sets the response header `x-microsoft-new-refresh-token` with the new refresh token
- Includes `res.headersSent` safety check

## Design

Uses a **callback pattern** to maintain separation of concerns:
- `graph-client.ts` has no knowledge of Express or `res` — it only calls `notifyTokenRefreshed()`
- `server.ts` binds the callback to `res.setHeader()` when setting up the request context
- The header is set **during** tool execution (before `StreamableHTTPServerTransport` writes the response), so it is included in the HTTP response

```
graph-client.ts                request-context.ts              server.ts
     |                               |                           |
     | 401 → refresh                 |                           |
     | new RT obtained               |                           |
     |-- notifyTokenRefreshed(RT) -->|                           |
     |                               |-- onTokenRefreshed(RT) -->|
     |                               |     res.setHeader(...)    |
     |                               |                           |
     | retry with new AT             |                           |
     |                    [transport writes response with header] |
```

## Clients

After this fix, HTTP mode clients should check for the `x-microsoft-new-refresh-token` response header on every `/mcp` response. If present, the client should update its stored refresh token with the new value.

## Tests

Added 10 new tests (all passing):
- `src/__tests__/request-context.test.ts` — callback invocation, edge cases (no callback, outside context, multiple calls)
- `src/__tests__/graph-client-refresh.test.ts` — 401 retry with callback, no 401 case, no refresh_token in response, missing callback

## Possibly related

- #287 — "Need to reregister OWUI client after X amount of time" may be a symptom of this issue

### DIFF
--- a/src/__tests__/graph-client-refresh.test.ts
+++ b/src/__tests__/graph-client-refresh.test.ts
@@ -1,0 +1,187 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { requestContext } from '../request-context.js';
+
+// Mock logger
+vi.mock('../logger.js', () => ({
+  default: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  },
+}));
+
+// Mock cloud-config
+vi.mock('../cloud-config.js', () => ({
+  getCloudEndpoints: () => ({
+    graphApi: 'https://graph.microsoft.com',
+    authority: 'https://login.microsoftonline.com',
+  }),
+}));
+
+// Mock microsoft-auth
+const mockRefreshAccessToken = vi.fn();
+vi.mock('../lib/microsoft-auth.js', () => ({
+  refreshAccessToken: (...args: any[]) => mockRefreshAccessToken(...args),
+}));
+
+// Mock toon
+vi.mock('@toon-format/toon', () => ({
+  encode: (data: any) => JSON.stringify(data),
+}));
+
+// Mock auth manager
+const mockAuthManager = {
+  getToken: vi.fn().mockResolvedValue('mock-access-token'),
+};
+
+const mockSecrets = {
+  clientId: 'test-client-id',
+  tenantId: 'test-tenant-id',
+  clientSecret: 'test-client-secret',
+  cloudType: 'global' as const,
+};
+
+// We need to import GraphClient after mocks are set up
+const { default: GraphClient } = await import('../graph-client.js');
+
+describe('GraphClient refresh token notification', () => {
+  let graphClient: InstanceType<typeof GraphClient>;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    graphClient = new GraphClient(mockAuthManager as any, mockSecrets);
+  });
+
+  it('should call notifyTokenRefreshed when 401 triggers a token refresh', async () => {
+    const onTokenRefreshed = vi.fn();
+
+    // First call returns 401, second returns 200
+    const mockFetch = vi
+      .spyOn(globalThis, 'fetch')
+      .mockResolvedValueOnce(
+        new Response('Unauthorized', { status: 401, statusText: 'Unauthorized' })
+      )
+      .mockResolvedValueOnce(new Response(JSON.stringify({ value: 'ok' }), { status: 200 }));
+
+    mockRefreshAccessToken.mockResolvedValueOnce({
+      access_token: 'new-access-token',
+      refresh_token: 'new-refresh-token-v2',
+      token_type: 'Bearer',
+      scope: 'User.Read',
+      expires_in: 3600,
+    });
+
+    await requestContext.run(
+      {
+        accessToken: 'expired-access-token',
+        refreshToken: 'old-refresh-token',
+        onTokenRefreshed,
+      },
+      async () => {
+        await graphClient.makeRequest('/me');
+      }
+    );
+
+    expect(onTokenRefreshed).toHaveBeenCalledWith('new-refresh-token-v2');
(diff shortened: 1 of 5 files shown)


eirikb:
This has been draft for a while now. Closing. Feel free to repoen :slightly_smiling_face:

PR #351, opened 2026-04-03
feat: add --slim mode minimal schema to reduce ai context flood with JSON string coercion for MCP clients

EurofaseDavid:
## Summary

- Introduces a `--slim` flag that registers all tools with minimal schemas (name + description only) to reduce AI model context overhead by ~8x
- Adds a `get-tool-schema` meta-tool for on-demand parameter lookup, now with **full body schema introspection** — exposes enum values, nested object shapes, array item types, and descriptions so callers know exactly what values are valid without consulting external docs
- Fixes a compatibility issue where MCP clients (including Claude Code) serialize the `parameters` field as a JSON string rather than a parsed object; `z.preprocess` is added to auto-parse JSON strings before Zod validation, preventing silent tool call failures (empty request bodies, unsubstituted path parameters) that manifested as 400/500 Graph API errors

## Root cause: auto-correct wrapping bug triggered by schema validation failure

The following bug chain was discovered during testing of `send-mail` in slim mode:

1. `get-tool-schema` previously returned only `name`, `placement`, and `required` — no enum values or nested types
2. Without schema guidance, the caller passes `contentType: "Text"` (PascalCase) — a natural assumption, but the Zod schema uses `z.enum(["text", "html"])` (lowercase only)
3. `send_mail_Body.safeParse({ Message: { body: { contentType: "Text", ... }, ... } })` **fails** due to the enum mismatch
4. The auto-correct logic in `executeGraphTool` catches the failure and retries by wrapping the value: `{ body: { Message: {...} } }`
5. This wrapped form passes Zod validation — `Message` is optional (`.partial()`), and `.passthrough()` allows the extra `body` key through
6. The HTTP body sent to Graph API is `{ "body": { "Message": {...} } }` — no top-level `Message` field
7. Graph API returns `400: One or more parameters of the operation 'sendMail' are missing from the request payload. The missing parameters are: Message`
8. Earlier attempts with no body at all produced `500: ErrorInternalServerError` — the same root cause manifesting differently depending on how the parameters were passed

**The fix:** `get-tool-schema` now exposes the full body schema so callers see `enum: ["text", "html"]` directly, eliminating the guesswork that triggers the validation failure and the subsequent wrapping misbehaviour.

Note: the auto-correct wrapping logic itself (step 4) is a separate latent bug — it can silently corrupt the request body whenever Zod validation fails for any reason. This is worth addressing independently.

## What `get-tool-schema` now returns for Body parameters

Body parameters now include a `schema` field with full structure, e.g. for `send-mail`:

```json
{
  "name": "body",
  "placement": "Body",
  "schema": {
    "type": "object",
    "properties": {
      "Message": {
        "type": "object",
        "properties": {
          "body": {
            "type": "object",
            "properties": {
              "contentType": { "type": "string", "enum": ["text", "html"] },
              "content": { "type": "string" }
            }
          },
          "subject": { "type": "string" },
          "toRecipients": { "type": "array" }
        }
      }
    }
  }
}
```

## Implementation

Adds `zodSchemaToJson()` — a recursive Zod introspection utility that walks `schema._def` to produce a JSON Schema-compatible representation:
- `ZodEnum` → `{ type: "string", enum: [...] }`
- `ZodObject` → `{ type: "object", properties: {...}, required: [...] }`
- `ZodArray` → `{ type: "array", items: {...} }`
- `ZodOptional`/`ZodNullable`/`ZodDefault` → unwrapped transparently
- `ZodLazy` → depth-limited to prevent infinite recursion
- Descriptions from `.describe()` preserved throughout

## Files changed

- `src/cli.ts`: add `--slim` CLI flag and `CommandOptions` property
- `src/server.ts`: slim mode branching, MCP server instructions, logging
- `src/graph-tools.ts`: `registerSlimTools()`, `isZodOptional()`, `zodSchemaToJson()`, `z.preprocess` fix
- `CHANGELOG.md`: full root-cause analysis and usage documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,199 @@
+# Changelog
+
+## [Unreleased] — Slim Mode Feature Branch
+
+### Overview
+
+This change set introduces a `--slim` operating mode for the ms-365-mcp-server. The feature was developed to solve a fundamental context-window problem when running the server with AI models such as Claude Code: the default full-registration mode exposes 200+ tools, each with complete Zod input schemas, which consumes the entire model context window and leaves the AI unable to reason or respond. Slim mode reduces context overhead by approximately 8× while retaining full tool functionality via an on-demand schema lookup mechanism.
+
+A subsequent fix was required after observing that Claude Code (following an internal update) serializes MCP tool argument values as JSON strings rather than parsed JSON objects. This caused all slim-mode tool calls to silently fail because Zod's `z.record(z.any())` rejects string input. A `z.preprocess` step was added to auto-parse JSON strings before Zod validation.
+
+---
+
+### Problem This Resolves
+
+**Without slim mode**, the server registers each Microsoft Graph API tool with its full Zod schema. An `npm run generate` produces 200+ tools. When an AI model connects, the complete tool list (with all parameter schemas) is injected into the model's context window. For Claude Code this consumed the entire context, making the server unusable in practice.
+
+**The slim mode solution** registers every tool with a single opaque input field (`parameters: Record<string, any>`), reducing the per-tool schema to name + description only. A dedicated `get-tool-schema` meta-tool allows the model to retrieve the full parameter list for any individual tool on demand, before calling it.
+
+**The string-coercion fix** addresses a breaking change in Claude Code's MCP client that began sending the `parameters` field value as a serialized JSON string (e.g., `'{"messageId":"..."}'`) rather than a parsed object. Without the fix, Zod's `z.record(z.any())` returns an `invalid_type` error ("Expected object, received string"), causing 100% of slim-mode tool calls to fail silently — path parameters are never substituted into URLs, request bodies are never sent, and Graph API returns 400 or 500 errors with no useful diagnostic.
+
+---
+
+### Files Changed
+
+#### `src/cli.ts`
+
+Added the `--slim` flag to the CLI option parser.
+
+**Before:**
+```ts
+.option('--discovery', 'Enable runtime tool discovery and loading (experimental feature)')
+.option('--cloud <type>', ...)
+```
+
+**After:**
+```ts
+.option('--discovery', 'Enable runtime tool discovery and loading (experimental feature)')
+.option(
+  '--slim',
+  'Register all tools with minimal schemas (name+description only). Use get-tool-schema for parameter details. Reduces context by ~8x vs full mode.'
+)
+.option('--cloud <type>', ...)
+```
+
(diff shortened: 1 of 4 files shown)


EurofaseDavid:
Context with --slim: ~26114 tokens, --discovery: ~500.  however claude continually fails to figure out how to use the tools, it can search and find a tool.  it can verify the email box, but can't list or send mail and gave up trying to figure out how to use the tools.  what worked with --slim was that it knew to use get-tool-schema and what each tool was for and the body schema provided on demand.   i cannot use --discovery to do anything.  if you can add a helper like get-tool-schema that may work.  unless there's something else I need to do.

i just updated this pull request with an improvement to the schema

eirikb:
The idea was that search-tools would inform the agent about the schema. No idea if it actually works the way it should.
It is a known pattern, often called dynamic tool loading, and now some agents will even do this automatically.

This is how it looks when I test:
image

And it also struggles, but eventually get it correct:
image

So it might be that we need to adjust `--discovery` instead?  It is a known MCP pattern and it should work.

But you think we should have both `--discovery` and `--slim`?  
If slim works better for you now, could that simply be because --discovery doesn't return the schema correctly?
If it did return the schema it should work just as well, with much less tokens?

EurofaseDavid:
Ideally the --discovery is the minimal tokens and would work.  discovering the correct format is critical if the full schema was returned.  reading seems to work much better than sending.  the mcp from claude/microsoft is read only and works fine for reading, i want this one to be able to send/draft.

eirikb:
I've been thinking about this PR. --discovery is obviously not 100% at the moment, and from a users perspective it's just one more optional flag to test out - so I think we should merge this.

But some things first:

  I'm happy to accept --slim as an experimental mode (like --discovery). A few things that need addressing
   before merge though:

  1. There is a conflict with #366 since I just merged that, check out `buildMcpServerInstructions` (instead of hard coded strings in server.ts)
  2. Drop the CHANGELOG
  3. Add tests
  4. I see duplicated code (parse-teams-url)
  5. Decide what to do about llmTips. They were originally meant to be "in addition" to spec, not as alternative prosa version. That has kind of become opposite because of LLM-based PRs, but in the future it might be fully cleaned up - meaning perhaps we should not include it?

eirikb:
Hi. Just a note, I've tried to improve --discovery

It didn't return spec in search at all - as it shouldn't (imagine multiple hits) - now it has a new tool.
So new tool + better search (not misses, and not only because of 4.7 :laughing: ):

image

eirikb:
Also, I think --discovery has gotten way better now! Could be this isn't needed, the benefit of a good --discovery outweighs having too many options.

PR #366, opened 2026-04-06
feat: add MCP server instructions and context handling

jmleksan:
## Summary

Adds MCP `initialize.instructions` for all server modes with specific instructions based on MCP mode like discovery or org mode. 

## Motivation

Hosts that honor MCP server instructions can inject a short usage guide for the model. Standard mode previously had no instructions; discovery-only text didn’t cover normal tool listings. This uses one shared baseline and appends discovery-specific workflow only when needed.

This was in response to a comment made on one of my PR's about such a fix. I thought i would give it a try to take some load off of your shoulders. Feel free to make any changes:  https://github.com/Softeria/ms-365-mcp-server/pull/349#issuecomment-4184110434

## Changes

- **`src/graph-tools.ts`:** `buildMcpServerInstructions({ discovery, orgMode, readOnly, multiAccount })`, `McpInstructionsContext`, general instructions + `DISCOVERY_MODE_INSTRUCTIONS_ADDON`.
- **`src/server.ts`:** `createMcpServer()` always passes `{ instructions: buildMcpServerInstructions(...) }` to `new McpServer(...)`.
- **`test/mcp-instructions.test.ts`:** Tests for standard vs discovery text and read-only.

### DIFF
--- a/src/mcp-instructions.ts
+++ b/src/mcp-instructions.ts
@@ -1,0 +1,39 @@
+/** Shared context for MCP `initialize.instructions` (hosts that forward it to the model). */
+export type McpInstructionsContext = {
+  orgMode: boolean;
+  readOnly: boolean;
+  multiAccount: boolean;
+};
+
+function buildGeneralMcpInstructions(opts: McpInstructionsContext): string {
+  const parts = [
+    'Microsoft 365 MCP exposes Microsoft Graph through MCP tools. Use each tool name, description, and parameter schema as the source of truth.',
+    'Microsoft Graph OData: do not combine $filter with $search on the same request. For lists, prefer modest $top (or top) and $select; avoid very large pages unless the user needs them.',
+    'Mail and message $search uses KQL; the $search query parameter value must be double-quoted per Graph (see search-query-parameter in Microsoft Graph docs).',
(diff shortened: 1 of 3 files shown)


eirikb:
This looks really good!

A couple of things:
  1. Do you think we should go through the `endpoints.json`-file in this same PR and trim down on now possible redundant instructions?
  2. Are you sure the instruction builder should be in graph-tools.ts? Perhaps a new file would be better?

I guess 2 is nitpicking to don't mind that, but 1 is important.
To be honest I think a lot of the `llmTip`s are just LLMs inserting LLM things just because it can (while LLMs already know these things, or it is already in spec).

jmleksan:
1. My main concern with that is that some clients may not support MCP instructions. Like OpenWebUI, which I use, currently does not although I added a patch for it to accept instructions. Therefore, I don't want to remove a bunch of stuff and break it for users who cannot take advantage of the new functionality yet.
2. I actually like that nitpick so I added it.

eirikb:
Agreed! Thanks :)

eirikb:
If we are to trim the `llmTip`s I think we should rename the property from llmTip to something else as well.
It made a lot of sense when it was made, but now in hindsight it seems when an LLM see the property it can't hold back and has to throw lots of text in there just because it can :sweat_smile:

PR #370, opened 2026-04-10
fix: persist rotated refresh tokens to prevent session disconnects

### LINKED ISSUE #336: Bug: Server-side 401 retry silently consumes refresh token without returning new one to client (token rotation breakage)

fabiankka:
## Summary

Fixes #336

When a Microsoft Graph API call returns `401`, the server refreshes the access token using the client-provided refresh token and retries the request. Due to Microsoft's [refresh token rotation](https://learn.microsoft.com/en-us/entra/identity-platform/refresh-tokens#refresh-token-lifetime), the old refresh token is invalidated and a new one is issued.

**The problem:** The new refresh token is silently discarded. The client (e.g. claude.ai) still holds the old, now-invalid refresh token. The next time the client tries to refresh, Microsoft rejects it — the session disconnects.

**The fix:** Server-side token persistence. After a successful 401 retry, the rotated tokens are stored in a file-backed cache. When the client later calls the `/token` endpoint with its stale refresh token, the server transparently substitutes the latest valid one from the store.

## What this fixes

Active sessions that span multiple token lifetimes (~1h per Azure AD access token). Without this fix, the first 401 retry consumes the refresh token and subsequent refreshes fail. With this fix, sessions survive indefinitely during active use.

## What this does NOT fix

Overnight disconnects where Azure AD itself expires the refresh token due to inactivity. That is a separate issue requiring a different architectural approach (e.g. server-side session management decoupled from Azure AD token lifetime).

### `src/token-store.ts` (new file, 83 lines)
- File-backed token store mapping sessions to their latest Azure AD tokens
- Sessions are keyed by refresh token prefix (stable across rotations)
- Automatic cleanup of sessions older than 7 days
- Atomic writes (write-to-tmp + rename)

### `src/graph-client.ts` (+10 lines)
- After a successful 401 retry with token refresh:
  - Updates the `RequestContext` so subsequent calls in the same request chain use the new tokens
  - Persists the rotated tokens to the store via `storeTokens()`

### `src/server.ts` (+19 lines)
- `/token` endpoint (`refresh_token` grant): checks the store for a more recent refresh token before calling Azure AD, and persists the result
- Token store initialized at server startup
- Hourly cleanup of expired sessions

## Design decisions

- **File-backed store, not in-memory only** — survives container restarts (important for Docker deployments with auto-updates via Watchtower etc.)
- **Session key = first 16 chars of refresh token** — stable identifier that doesn't change across rotations. Both the original and rotated token prefix are indexed for lookup
- **No new dependencies** — uses only `node:fs` and `node:path`
- **Backward compatible** — if no rotation occurs, behavior is identical to before. The store is only consulted when a refresh token is presented to `/token`
- **Transparent to clients** — no protocol changes, no new headers, no client-side changes needed. Works with claude.ai, Claude Desktop, ChatGPT, and any other MCP client

## How to verify

1. Connect from claude.ai and use any tool (triggers initial auth)
2. Wait >1h for the access token to expire
3. Use another tool — this triggers a 401 retry with token refresh
4. Check logs for `Token store: session updated`
5. Wait another >1h and repeat
6. Previously this would disconnect; with the fix, the session persists

## Relationship to #337

PR #337 (draft) takes a different approach: returning the new refresh token via a custom HTTP response header (`x-microsoft-new-refresh-token`). That approach requires the client to read and process custom headers, which claude.ai and most MCP clients do not do. This PR solves the problem server-side, requiring no client changes.

### DIFF
--- a/src/graph-client.ts
+++ b/src/graph-client.ts
@@ -5,6 +5,7 @@ import { encode as toonEncode } from '@toon-format/toon';
 import type { AppSecrets } from './secrets.js';
 import { getCloudEndpoints } from './cloud-config.js';
 import { getRequestTokens } from './request-context.js';
+import { storeTokens } from './token-store.js';
 
 interface GraphRequestOptions {
   headers?: Record<string, string>;
@@ -66,6 +67,15 @@ class GraphClient {
         // Token expired, try to refresh
         const newTokens = await this.refreshAccessToken(refreshToken);
         accessToken = newTokens.accessToken;
+
+        // Persist rotated tokens so subsequent requests and client refreshes use the new RT
+        if (newTokens.refreshToken) {
+          if (contextTokens) {
+            contextTokens.accessToken = newTokens.accessToken;
+            contextTokens.refreshToken = newTokens.refreshToken;
+          }
+          storeTokens(refreshToken, newTokens.accessToken, newTokens.refreshToken);
+        }
 
         // Retry the request with new token
         response = await this.performRequest(endpoint, accessToken, options);
--- a/src/server.ts
+++ b/src/server.ts
@@ -18,6 +18,12 @@ import type { CommandOptions } from './cli.ts';
 import { getSecrets, type AppSecrets } from './secrets.js';
 import { getCloudEndpoints } from './cloud-config.js';
 import { requestContext } from './request-context.js';
+import {
+  initTokenStore,
+  getLatestRefreshToken,
+  storeTokens,
+  cleanupTokenStore,
+} from './token-store.js';
 import crypto from 'node:crypto';
 
 /**
@@ -425,6 +431,13 @@ class MicrosoftGraphServer {
             const clientId = this.secrets!.clientId;
             const clientSecret = this.secrets?.clientSecret;
 
+            // Use the latest refresh token from store (may have been rotated by a 401 retry)
+            const clientRT = body.refresh_token as string;
+            const latestRT = getLatestRefreshToken(clientRT);
+            if (latestRT !== clientRT) {
+              logger.info('Refresh endpoint: Using rotated refresh token from store');
+            }
+
             // Log whether using public or confidential client
             if (clientSecret) {
               logger.info('Refresh endpoint: Using confidential client with client_secret');
@@ -433,12 +446,18 @@ class MicrosoftGraphServer {
             }
 
             const result = await refreshAccessToken(
-              body.refresh_token as string,
+              latestRT,
               clientId,
               clientSecret,
               tenantId,
               this.secrets!.cloudType
             );
+
+            // Persist the new tokens
+            if (result.refresh_token) {
+              storeTokens(clientRT, result.access_token, result.refresh_token);
+            }
+
             res.json(result);
           } else {
             res.status(400).json({
@@ -570,6 +589,14 @@ class MicrosoftGraphServer {
       app.get('/', (req, res) => {
         res.send('Microsoft 365 MCP Server is running');
       });
+
+      // Initialize token rotation store (persists Azure AD tokens across 401 retries)
+      const tokenStoreDir =
+        process.env.MS365_MCP_TOKEN_STORE_DIR || new URL('..', import.meta.url).pathname;
+      initTokenStore(tokenStoreDir);
+
+      // Cleanup expired sessions every hour
+      setInterval(cleanupTokenStore, 60 * 60 * 1000);
 
       if (host) {
         app.listen(port, host, () => {
--- a/src/token-store.ts
+++ b/src/token-store.ts
@@ -1,0 +1,83 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import logger from './logger.js';
+
+interface StoredTokens {
+  accessToken: string;
+  refreshToken: string;
+  updatedAt: number;
+}
+
+const tokenMap = new Map<string, StoredTokens>();
+let storePath: string | undefined;
+
+export function initTokenStore(dataDir?: string): void {
+  if (!dataDir) return;
+  storePath = path.join(dataDir, 'token-cache.json');
+  try {
+    const data = JSON.parse(fs.readFileSync(storePath, 'utf-8'));
+    for (const [k, v] of Object.entries(data)) {
+      tokenMap.set(k, v as StoredTokens);
+    }
+    logger.info(`Token store loaded: ${tokenMap.size} sessions`);
+  } catch {
+    logger.info('Token store: starting fresh');
+  }
+}
+
+function save(): void {
+  if (!storePath) return;
+  try {
+    const tmp = storePath + '.tmp';
+    fs.writeFileSync(tmp, JSON.stringify(Object.fromEntries(tokenMap)));
+    fs.renameSync(tmp, storePath);
+  } catch (e) {
+    logger.error('Token store save failed:', e);
+  }
+}
+
+/**
+ * Derive a stable session key from the refresh token prefix.
+ * The first refresh token in a session establishes the key;
+ * subsequent rotations update the stored tokens but keep the same key.
+ */
+function sessionKey(refreshToken: string): string {
+  return refreshToken.slice(0, 16);
+}
+
+export function storeTokens(
+  originalRefreshToken: string,
+  accessToken: string,
+  refreshToken: string
+): void {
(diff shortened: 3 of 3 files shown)


eirikb:
Is this for HTTP mode (--http)?

mleitnercom:
Yes, HTTP mode. Deployed this PR, connected via ChatGPT auto-discovery — OAuth flow, two-leg PKCE, Graph calls all work. The 401-retry rotation path hasn't fired yet (tokens still fresh). Will report back in 1-2 weeks on long-run stability. Happy to add a test for `storeTokens()` / `getLatestRefreshToken()` if that'd help.

eirikb:
Ok I will take a look.  But honestly I don't quite understand,  when in http mode we can't store tokens on the server. We don't do that, and we should not do that

mleitnercom:
Context for why this matters to me: on the hosted `@softeria/ms-365-mcp-server` (HTTP mode, Cloudflare tunnel, Azure consumer app), I've had Claude.ai and ChatGPT connectors losing their session every few hours. Re-auth every time. The symptom matches this PR's target: once a server-side 401 retry rotates the RT at Microsoft, the client's stored RT is quietly dead, the next `/token` call fails, the connector disconnects.

That's what pulled me toward this PR.

On what's stored: one entry per session-key (first 16 chars of initial RT), `{access_token, refresh_token, updatedAt}`, 7-day TTL, atomic writes. Narrow by design — sole purpose is to let the next `/token` call from a stale-RT client be served from the store instead of failing.

On #337's alternative (send new RT back via response header): I honestly don't know whether the hosted Claude.ai and ChatGPT connectors read a custom header to update their stored RT. If they do and the server just isn't sending it, that's the exact fix I need. If they don't, #337 can't reach the big hosted clients and some server-side memory is unavoidable. Is there an MCP spec hook for server-initiated RT rotation I'm missing?

Happy to shape this differently, or pivot to #337, depending on what's reachable.

mleitnercom:
Important clarification after adding auth-middleware logging to both a stock 0.79.6 instance and this PR's branch.

Claude.ai never sends `x-microsoft-refresh-token` to our server. Every request: `has_rt=false rt_len=0`. So the `if (response.status === 401 && refreshToken)` branch in `graph-client.ts` (which this PR extends) can never fire for Claude.ai. ChatGPT behaves the same on a quick check.

Timeline observed over ~4 hours with stable debug logs on both endpoints:

- Initial OAuth at T=0, tokens work for everything
- ~60 min later: Microsoft starts rejecting writes on MSA compact tokens at `/me/sendMail` and similar action endpoints with `IDX14100 JWT not well formed`
- Eventually reads fail too on the same AT (`token_hash` unchanged, globally invalidated)
- Claude.ai never attempts to refresh. It just keeps re-sending the same stale AT. No `/token` call to our authorization server ever logged
- Full reconnect in the Claude.ai UI is the only thing that restores function — for another ~60 min

So the dominant hosted-client failure mode isn't what this PR addresses. The server side doesn't get a chance to refresh: the RT isn't reaching us, and Claude.ai isn't asking our `/token` endpoint either. Neither this PR nor #337 reach the code path.

Not arguing against #370 — the scenario it targets is real and the fix looks clean for clients that do send the RT. Just making sure the scope is clear. If it would be useful, I can open a separate issue documenting the hosted-client-every-hour failure mode; feels like a different axis (server-authoritative token state would be needed for a full fix).

eirikb:
Please see #406 . It's a different way of solving this, while still keeping things stateless.  
There were severe bugs in the auth handling, which I hope that PR solves.  
Please feel free to test, it might help.  

I understand that some clients don't support this now, and that storing tokens server-side would mitigate this, but my gut feeling is screaming at me to keep http mode stateless.
I also understand that single-user http mode is a different matter, but opening up for this feature can have unthinkable consequences.  

Solving these issues on the client is the correct solution, e.g., https://github.com/anthropics/claude-code/issues/46328
 
I'm going to close this for now.

PR #372, opened 2026-04-10
feat: readOnly has been enhanced. a readOnly attribute has been introduce so that any Graph API endpoint is actually reading the data although the METHOD is POST.

zahangirbu:
feat: readOnly has been enhanced. a readOnly attribute has been introduce so that any Graph API endpoint is actually reading the data although the METHOD is POST.

There are some Graph API POST endpoints that only reads the data, e.g.,  /me/calendar/getSchedule, /me/findMeetingTimes. Although these are POST endpoints but there scopes are read only. Also test cases are updated for this scenario.


eirikb:
Hi there. This PR is good, the read only feature would be nice. But there is a docker-build file there, could you remove it?

zahangirbu:
hi @eirikb, thanks for your feedback. I have removed docker-build file.

PR #374, opened 2026-04-10
Add hosted-content endpoints for chat and channel messages

tbarsness:
## Summary

Adds four new Graph endpoints to `src/endpoints.json` so MCP clients can read Teams hosted-content references (the inline images, code snippets, and similar resources embedded in chat and channel message bodies):

- `list-chat-message-hosted-contents` — `GET /chats/{chat-id}/messages/{chatMessage-id}/hostedContents`
- `get-chat-message-hosted-content` — `GET /chats/{chat-id}/messages/{chatMessage-id}/hostedContents/{chatMessageHostedContent-id}/$value`
- `list-channel-message-hosted-contents` — `GET /teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/hostedContents`
- `get-channel-message-hosted-content` — `GET /teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/hostedContents/{chatMessageHostedContent-id}/$value`

The two `get-*` variants set `returnDownloadUrl: true`, matching the existing pattern already used by `get-my-profile-photo` for `/me/photo/$value`.

## Motivation

Today I hit a wall trying to read a screenshot a colleague had posted inline in a Teams 1:1 chat. The message body retrieved via `list-chat-messages` / `get-chat-message` contained an `` reference, but there was no exposed tool to actually fetch the bytes. The MCP server already holds an auth token that can authenticate that endpoint, but the call surface wasn't wired up.

With these endpoints exposed, an LLM client can:
1. Read a chat message, notice the hosted-content URL in the body
2. Call `list-chat-message-hosted-contents` to enumerate attachments (or extract the ID from the img src URL directly)
3. Call `get-chat-message-hosted-content` to retrieve the actual image/png bytes

## Scopes

Scopes match the existing chat/channel message tools in `endpoints.json`:

- Chat variants use `ChatMessage.Read` (same as `list-chat-messages`)
- Channel variants use `ChannelMessage.Read.All` (same as `list-channel-messages`)

So no additional Azure app-registration consent is required for existing installs — the new tools work with the permissions users already granted.

## Test plan

- [x] `npm run generate` succeeds and the 4 new aliases appear in `src/generated/client.ts`
- [x] `npm run verify` passes end to end (all 117 tests green, lint + format + build clean)
- [x] Manually smoke-tested `list-chat-message-hosted-contents` + `get-chat-message-hosted-content` against a real Teams 1:1 message containing an inline screenshot; returned a valid PNG

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -945,6 +945,21 @@
     "workScopes": ["ChatMessage.Read"]
   },
   {
+    "pathPattern": "/chats/{chat-id}/messages/{chatMessage-id}/hostedContents",
+    "method": "get",
+    "toolName": "list-chat-message-hosted-contents",
+    "workScopes": ["ChatMessage.Read"],
+    "llmTip": "Lists hosted-content references (inline images, code snippets) attached to a Teams chat message. Returns an array of { id, contentType }. Use get-chat-message-hosted-content with the id to download the bytes. Hosted content IDs can also be parsed from the <img src> URL in the message body between /hostedContents/ and /$value."
+  },
+  {
+    "pathPattern": "/chats/{chat-id}/messages/{chatMessage-id}/hostedContents/{chatMessageHostedContent-id}/$value",
+    "method": "get",
+    "toolName": "get-chat-message-hosted-content",
+    "workScopes": ["ChatMessage.Read"],
+    "returnDownloadUrl": true,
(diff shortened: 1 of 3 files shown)


eirikb:
Thanks! This was a nice PR.

One thing; I believe the content will go through LLM? That can be interesting (and expensive).
But there really isn't any other option now.

I've mentioned it before that it would be nice with a helper-tool that the LLM could use for direct file download. Mail attachments need the same feature I believe.
So some time in the future perhaps :laughing:

PR #385, opened 2026-04-14
fix: restrict default CORS origin from wildcard to localhost

okapi-ca:
## Security fix — Critical severity (HTTP mode only)

**Problem**: The HTTP mode CORS default is `Access-Control-Allow-Origin: *`. This allows **any website** to make cross-origin requests to the MCP server, including:
- OAuth endpoints (`/authorize`, `/token`)
- MCP tool endpoints (`/mcp`)
- Discovery endpoints (`/.well-known/*`)

A malicious page visited by an authenticated user could invoke Graph API tools (read emails, send messages, access files) via the MCP server.

**Fix**: Change default from `*` to `http://localhost:3000`. Deployments needing broader access can set `MS365_MCP_CORS_ORIGIN` to their specific origin, or explicitly to `*` if they accept the risk.

**Impact**: HTTP mode only. Stdio mode is unaffected. Existing deployments that rely on `*` must set `MS365_MCP_CORS_ORIGIN=*` explicitly. This is a **breaking change** for HTTP mode deployments accessed from origins other than localhost.

## Test plan

- [x] `npm run verify` passes
- [ ] Verify localhost MCP client still works in HTTP mode
- [ ] Verify cross-origin requests from other origins are blocked unless `MS365_MCP_CORS_ORIGIN` is set

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -169,7 +169,7 @@ class MicrosoftGraphServer {
       app.use(express.urlencoded({ extended: true }));
 
       // Add CORS headers for all routes
-      const corsOrigin = process.env.MS365_MCP_CORS_ORIGIN || '*';
+      const corsOrigin = process.env.MS365_MCP_CORS_ORIGIN || 'http://localhost:3000';
       app.use((req, res, next) => {
         res.header('Access-Control-Allow-Origin', corsOrigin);
         res.header('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, OPTIONS');


eirikb:
I just wonder - CORS for whom? What browser?

okapi-ca:
Fair question. The scenario is: the MCP server runs in HTTP mode on `localhost:3000`. The user opens a malicious or compromised website in their regular browser (Chrome, Firefox, any). With `Access-Control-Allow-Origin: *`, that website's JavaScript can make `fetch()` calls to `http://localhost:3000/mcp` — and the browser will happily send them because CORS says "allow all origins."

Since the user is already authenticated (token stored server-side), those requests execute with the user's Graph permissions. The malicious site could read emails, send messages, list files — all silently.

With the fix, only `localhost` origin is allowed by default, so browser-based cross-origin requests from external sites are blocked. Anyone running a non-localhost frontend can set `MS365_MCP_CORS_ORIGIN` to their specific origin.

That said, if you consider this an acceptable risk for the use case (developer tool, localhost only), I understand. But the default `*` is unusually permissive for a server that holds Graph API tokens.

eirikb:
We don't store tokens for http mode, so it wouldn't work

eirikb:
I merged it, but I honestly don't think this is a security issue. It's AI making things up.
But I don't see any harm in adding it, and the existing thing was also added by some AI making things up PR :smiling_face_with_tear:

PR #392, opened 2026-04-16
fix: correct list-chats to use beta API for viewpoint, add llmTips

vongohren:
## Summary

- **`list-chats`**: Add `apiVersion: "beta"` so `viewpoint` (unread detection) works. Fix llmTip to use `$select` instead of `$expand` — viewpoint is a complex property, not a navigation property.
- **`list-chat-messages`**: Add llmTip guiding LLMs toward `$orderby=lastModifiedDateTime desc` with client-side date filtering for "what's new" queries.
- **`graph-client.ts` / `graph-tools.ts`**: Support `apiVersion` option to override the default v1.0 Graph API version per-request, driven by `apiVersion` in `endpoints.json`.
- **Tests**: Endpoint config invariant tests, apiVersion passthrough unit tests, and an LLM tip eval harness (`test/llm-tip-evals.ts`) that verifies llmTips actually guide Claude to correct tool calls.

## Context

`viewpoint` on `/me/chats` only works on the beta Graph API. The current code hardcodes `v1.0` in graph-client.ts, so any endpoint with `apiVersion: "beta"` in endpoints.json had no effect unless it was a beta-only endpoint registered separately at runtime. For `list-chats` (which exists in both v1.0 and beta), the v1.0 registration wins and the beta config is ignored — resulting in a 400 error when following the llmTip to expand viewpoint.

Also: `/chats/{chat-id}/messages/delta()` appears in Graph OData metadata but is not implemented (returns "Change tracking is not supported against microsoft.graph.chatMessage"). See [microsoftgraph/msgraph-metadata#607](https://github.com/microsoftgraph/msgraph-metadata/issues/607). This PR does not add that endpoint but notes it in the `list-chat-messages` llmTip as a known limitation.

## Test plan

- [x] Unit tests pass (`npm test` — 124 passing, 1 pre-existing failure in orphan check unrelated to this PR)
- [x] LLM tip evals pass (`npx tsx test/llm-tip-evals.ts` — 4/4)
- [x] Manually verified: `list-chats` with `$select=viewpoint` returns `viewpoint.lastMessageReadDateTime` on beta
- [x] Manually verified: `list-chat-messages` with `$orderby=lastModifiedDateTime desc` works for incremental chat scanning
- [ ] Reviewer: confirm `apiVersion` passthrough doesn't affect existing v1.0 endpoints (test covers this)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,6 +144,10 @@ dist
 # Server logs (may contain PII and token metadata)
 logs/
 
+# LLM eval ephemeral outputs
+test/llm-tip-eval-results.json
+test/.eval-tmp/
+
 openapi
 
 src/generated/client.ts
--- a/README.md
+++ b/README.md
@@ -589,6 +589,30 @@ The Key Vault integration uses `DefaultAzureCredential` from the Azure Identity 
 
 The Azure Key Vault packages (`@azure/identity` and `@azure/keyvault-secrets`) are optional dependencies. They are only loaded when `MS365_MCP_KEYVAULT_URL` is configured. If you don't use Key Vault, these packages are not required.
 
+## Testing
+
+### Unit tests
+
+```bash
+npm test
+```
+
+Includes endpoint config validation (`test/endpoints-validation.test.ts`) and graph-tools logic tests (`src/__tests__/graph-tools.test.ts`).
+
+### LLM tip evals
+
+Non-deterministic tests that verify llmTips in `endpoints.json` actually guide an LLM to make correct tool calls. Requires the `claude` CLI.
+
+```bash
+npx tsx test/llm-tip-evals.ts           # run all evals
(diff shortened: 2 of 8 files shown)


eirikb:
Hi. This is very interesting. I've considered beta to be up-and-in-the-future, but perhaps the future is now, and it's quite easy to add - except it's a bit more complicated: The spec generator points to v1. The endpoints you list out happen to be in v1, but it won't work for any other. So to make this complete I think we need to add version support in the generator.

eirikb:
Also it has conflicts, and doesn't build.

I totally agree that having beta support would be great, but it requires a bit more work to do properly - it must be part of the spec parser!

vongohren:
I agree, it did not become a priority for us, running it locally for now.
Will fix later at some point

PR #411, opened 2026-04-22
fix: send Origin header on /token to satisfy SPA redirect policy (code + refresh)

okapi-ca:
## Summary

When the app registration has the incoming `redirect_uri` registered as a **Single-Page Application**, Entra requires the server-side POST to `/oauth2/v2.0/token` to include an `Origin` header matching that redirect_uri's origin. Node's `fetch` does not send `Origin` by default, so the exchange fails with `AADSTS9002327` even though PKCE is correct.

This change derives the `Origin` header from the `redirect_uri` and attaches it to the `/token` request. Web-type redirects ignore the header; SPA-type redirects get unblocked.

## Motivation

Running this MCP server against a tenant with restrictive consent policies (common in enterprise M365 tenants) means the app registration often cannot rely on `web` redirects + `isFallbackPublicClient=true`:

- `web` redirect + no client secret + PKCE → `AADSTS7000218: client_secret required`. Entra still treats the app as confidential in that path despite `isFallbackPublicClient`.
- Adding a client secret is undesirable: it adds a rotation burden for an MCP server that is designed to work with end-user delegated tokens via PKCE, and it changes the DCR story for hosted deployments.
- **SPA** redirect_uri is the documented v2.0 path for pure-PKCE public clients and works without a secret.

Switching the redirect to SPA hits the next guard rail: `AADSTS9002327` when the server tries to exchange the code without `Origin`. That is what this PR fixes.

## Fix

```ts
// src/lib/microsoft-auth.ts — exchangeCodeForToken
const headers: Record<string, string> = {
  'Content-Type': 'application/x-www-form-urlencoded',
};
try {
  const redirectUrl = new URL(redirectUri);
  headers['Origin'] = redirectUrl.origin;
} catch {
  // redirect_uri is not a valid URL — omit Origin and let Entra decide
}
```

Two-line addition inside the existing helper. No other flows touched.

## Behavior matrix

| Redirect type | Origin header effect | Before this PR | After |
|---|---|---|---|
| `web` (confidential) | Ignored by Entra | Works (with secret) | Works (with secret) |
| `web` + `isFallbackPublicClient` (PKCE, no secret) | Ignored by Entra | Fails: AADSTS7000218 | Fails: AADSTS7000218 (unchanged — requires SPA redirect) |
| `spa` (PKCE, no secret) | Required | Fails: AADSTS9002327 | **Works** |
| `publicClient` (native/loopback) | N/A | Works | Works |

Only `spa` behavior changes, and from "broken" to "working".

## Follow-up (not in this PR)

`refreshAccessToken` on the same file would need the same Origin header for the SPA refresh-token flow to work. That path isn't exercised by the current HTTP OAuth proxy (refresh is driven by the MCP client since #406), so I'm leaving it for a separate PR once someone has a concrete repro.

## Test plan

- [x] `npm run verify` passes (lint + format + build + tests, 173/173)
- [x] Manual: end-to-end sign-in completes when redirect_uris are registered under the app's `spa` block (`claude.ai/api/mcp/auth_callback` → no `AADSTS9002327`)
- [x] Manual: existing `web`-redirect flow with client_secret still works (verified against `http://localhost:3000/oauth/callback`)

### DIFF
--- a/src/lib/microsoft-auth.ts
+++ b/src/lib/microsoft-auth.ts
@@ -147,7 +147,8 @@ export async function refreshAccessToken(
   clientId: string,
   clientSecret: string | undefined,
   tenantId: string = 'common',
-  cloudType: CloudType = 'global'
+  cloudType: CloudType = 'global',
+  origin?: string
 ): Promise<{
   access_token: string;
   token_type: string;
@@ -166,11 +167,21 @@ export async function refreshAccessToken(
     params.append('client_secret', clientSecret);
   }
 
+  // Refresh tokens issued via a SPA redirect carry a SPA marker; Entra rejects
+  // a server-side refresh without an Origin header matching an allowed SPA
+  // origin (same AADSTS9002327 class as the initial code exchange). Forward
+  // the incoming request's Origin when available so refresh works the same
+  // way SPA redemption does. Non-SPA deployments are unaffected.
+  const headers: Record<string, string> = {
+    'Content-Type': 'application/x-www-form-urlencoded',
+  };
+  if (origin) {
+    headers['Origin'] = origin;
+  }
+
   const response = await fetch(`${cloudEndpoints.authority}/${tenantId}/oauth2/v2.0/token`, {
     method: 'POST',
-    headers: {
-      'Content-Type': 'application/x-www-form-urlencoded',
-    },
+    headers,
     body: params,
   });
 
--- a/src/lib/microsoft-auth.ts
+++ b/src/lib/microsoft-auth.ts
@@ -104,11 +104,29 @@ export async function exchangeCodeForToken(
     params.append('code_verifier', codeVerifier);
   }
 
+  // When the app registration has the redirect_uri registered as a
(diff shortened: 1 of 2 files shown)


okapi-ca:
Adding a follow-up commit (3032444) that extends the same Origin-header fix to `refreshAccessToken`.

Refresh tokens issued through a SPA-redirect flow carry the SPA marker, so Entra rejects a server-side refresh without an Origin header with the same AADSTS9002327 — just a ~60 min later, silently killing the session at first access-token expiry. The fix mirrors the code-exchange path but forwards the incoming request's Origin (or Referer fallback) since the caller doesn't keep the original redirect_uri around at refresh time.

Validated end-to-end against a tenant with `spa` redirect URIs: initial sign-in and the first silent refresh both succeed now. Updated the PR title/description to reflect the broader scope (was: exchangeCodeForToken only). 173/173 tests pass.

eirikb:
Hmm.. One question - so now the redirect must be SPA?
So the old setup, and  the one in readme, is now wrong?

Because my test in claude code now looks like this:
image

Issue #298, opened 2026-03-24
Token cache desync between keytar and file storage causes repeated login prompts


mrdoro:
Every new MCP server session requires re-authentication via device code flow, even though a valid refresh token exists in the
  file-based cache (.token-cache.json).

  **Root Cause**

  There is a desync between the two token storage backends (keytar / file). The saveTokenCache() method writes to either keytar or
  the file — never both:

  // src/auth.ts lines 277-302
  async saveTokenCache(): Promise<void> {
      const cacheData = this.msalApp.getTokenCache().serialize();
      try {
          const kt = await getKeytar();
          if (kt) {
              await kt.setPassword(SERVICE_NAME, TOKEN_CACHE_ACCOUNT, cacheData);
              // ← returns here, file is NOT updated
          } else {
              fs.writeFileSync(cachePath, cacheData, { mode: 0o600 });
          }
      } catch (keytarError) {
          // fallback to file only on error
          fs.writeFileSync(cachePath, cacheData, { mode: 0o600 });
      }
  }

  Meanwhile, loadTokenCache() (lines 210-242) prioritizes keytar — if keytar returns any data, the file cache is skipped entirely:

  if (!cacheData && existsSync(cachePath)) {
      cacheData = readFileSync(cachePath, 'utf8');
  }

  This creates a scenario where:
  1. Session starts → loadTokenCache() reads stale tokens from keytar
  2. getToken() attempts silent refresh with an expired/revoked refresh token → fails
  3. User must re-authenticate via device code flow
  4. saveTokenCache() attempts to write to keytar — if the write silently fails (e.g. keychain locked, permissions issue, context
  difference), it falls back to the file
  5. File now has fresh tokens, but keytar still has stale data
  6. Next session → step 1 repeats

  In my case, keytar had tokens cached from November 2025 while the file had fresh tokens from today. Every new session loaded the
  stale keytar data and forced re-authentication.

  **Proposed Fix**

  saveTokenCache() should write to both keytar and file, so that even if one backend becomes stale, the other is up to date. At
  minimum, the file should always be written as a reliable fallback:

  async saveTokenCache(): Promise<void> {
      const cacheData = this.msalApp.getTokenCache().serialize();

      // Always write to file as reliable fallback
      const cachePath = getTokenCachePath();
      ensureParentDir(cachePath);
      fs.writeFileSync(cachePath, cacheData, { mode: 0o600 });

      // Also try keytar
      try {
          const kt = await getKeytar();
[...]

PR #422, opened 2026-04-25
fix: normalize keytar ESM/CJS interop so token persistence works on Node 24+

### LINKED ISSUE #418: keytar ESM import breaks token persistence on Node 24 (kt.setPassword is not a function)
## Summary
`dist/auth.js` imports keytar via `await import("keytar")` but expects methods at top level. Under Node 24.x on Linux, this yields a namespace object where `kt.setPassword` is undefined (functions are under `kt.default`), triggering the warning:

```
WARN: Keychain save failed, falling back to file storage: kt.setPassword is not a function
WARN: Keychain save failed for selected account, falling back to file storage: kt.setPassword is not a function
```

The file fallback works correctly, so login eventually succeeds — but the keychain path never does. Also surfaces confusion because the same cache file ends up storing both the "fallback" token and the selected account without the user intending filesystem storage.

## Repro
- Node v24.15.0, Linux, libsecret + gnome-keyring installed and running, D-Bus session active
- `npx -y @softeria/ms-365-mcp-server --login --org-mode` completes login but logs the warning twice and skips keychain

Confirmed keytar itself works when required directly (CommonJS): `require('keytar').setPassword` is a function. The issue is ESM/CJS interop — dynamic `import("keytar")` returns `{default: <the-module>, ...named-re-exports}`, and the code at `dist/auth.js:214, 238` calls `kt.setPassword` / `kt.getPassword` on the namespace object instead of `kt.default`.

## Suggested fix
Normalize the import:

```js
const mod = await import("keytar");
keytar = mod.default ?? mod;
```

Same pattern for getPassword / deletePassword call sites.

## Environment
- `@softeria/ms-365-mcp-server` 0.85.1
- Node v24.15.0
- Linux, libsecret-1-0 0.21.7-2, gnome-keyring 48.0-5

treerobin06:
## Summary

Closes #418.

`dist/auth.js` (and its source in `src/auth.ts:19`) imports keytar with
`keytar = await import('keytar')` and then calls `kt.setPassword(...)` directly.
Under Node 24+ that namespace object's top-level `setPassword` is `undefined` —
the actual functions live on `.default`. As a result every save throws
\`kt.setPassword is not a function\`, silently falls back to file storage, and
on systems where the keychain *is* reachable tokens never get persisted there.
For users running through `npx` on Node 24+ this manifests as **having to re-login
every session** even though logging in succeeds and the file fallback may or may
not survive depending on environment.

This PR normalizes the import so we always pick the real module shape:

\`\`\`ts
const mod = (await import('keytar')) as typeof import('keytar') & {
  default?: typeof import('keytar');
};
keytar = mod.default ?? mod;
\`\`\`

`mod.default ?? mod` keeps backward compatibility with older Node versions
and pure-CJS bundlers where methods sit on the namespace itself.

## Reproduction (independent confirmation of #418 on macOS / Node 25)

The original report (#418) was on Linux Node 24.15.0. I hit the same bug on
**macOS 15 / Node v25.8.1** with `@softeria/ms-365-mcp-server@0.87.0` —
confirming it is not Linux- or libsecret-specific.

\`\`\`bash
$ node --version
v25.8.1

$ cd node_modules/@softeria/ms-365-mcp-server
$ node --input-type=module -e "
const m = await import('keytar');
console.log('top setPassword:', typeof m.setPassword);
console.log('top getPassword:', typeof m.getPassword);
console.log('m.default:', typeof m.default);
console.log('m.default.setPassword:', typeof m.default?.setPassword);
console.log('keys:', Object.keys(m));
"
top setPassword: undefined          ← what the current code calls (throws)
top getPassword: function           ← partial-export, misleadingly works
m.default: object
m.default.setPassword: function     ← the real function
keys: [ 'default', 'getPassword', 'module.exports' ]
\`\`\`

After the patch, the same module exposes \`setPassword\`/\`getPassword\`/\`deletePassword\`
as functions and a real round-trip write/read/delete against the macOS Keychain
succeeds.

## Behavior change

- ✅ Node 24+ (and Node 25): keychain persistence now works as intended; users
  no longer hit \"Keychain save failed, falling back to file storage\" on every
  login.
- ✅ Older Node / CJS contexts: unchanged — \`mod.default\` is undefined there
  so the \`?? mod\` branch keeps the previous behavior.
- ✅ Environments without keytar installed (alpine etc.): unchanged — the
  outer \`try/catch\` still routes to file-only storage.

## Tests

- \`npm run generate && npm test\` → **173 passed (25 files)** locally on Node v25.8.1.
- \`npm run lint\` → no new warnings; only pre-existing ones.
- \`npx prettier --check src/auth.ts\` → clean.

## Related

- Fixes #418 (\"keytar ESM import breaks token persistence on Node 24\")
- Likely root cause for repeated-login symptom in #287, #298 (closed but
  similar shape)

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -16,7 +16,14 @@ async function getKeytar() {
   }
   if (keytar === null) {
     try {
-      keytar = await import('keytar');
+      // Normalize ESM/CJS interop: under Node 24+ `await import('keytar')` returns a
+      // namespace object whose top-level `setPassword` is undefined (functions live on
+      // `.default`). On older Node and pure CJS, methods live on the namespace itself.
+      // Falling back to the namespace keeps backward compatibility. See issue #418.
+      const mod = (await import('keytar')) as typeof import('keytar') & {
+        default?: typeof import('keytar');
+      };
+      keytar = mod.default ?? mod;
       return keytar;
     } catch (error) {
       logger.info('keytar not available, using file-based credential storage');


eirikb:
LGTM :+1:

PR #425, opened 2026-04-25
feat: OneDrive sharing-link, copy, and preview action endpoints

BruceGao20:
## Summary

Follow-up to #423 in the same vein — adding Graph drive-item *action* endpoints that round out the OneDrive surface for programmatic workflows. Each fills a real gap that previously required workarounds (download-and-reupload for copies, share-with-invite for link generation, etc.).

### What's in this PR

- **`create-drive-item-share-link`** — POST `/drives/{id}/items/{id}/createLink`. Generates a shareable link (`view` / `edit` / `embed`; `anonymous` / `organization` / `users` scope) **without** sending an email invitation. Complements the existing `share-drive-item` (which uses `/invite` and triggers a OneDrive notification email). Use this when you only need a URL to paste into a doc, email, or chat — common for programmatic flows where the caller controls the messaging.

- **`copy-drive-item`** — POST `/drives/{id}/items/{id}/copy`. Asynchronously copy a file or folder to a new location and/or name via Graph's server-side copy. Returns `202 Accepted` + `Location` header pointing at a monitor URL for the async job. Ideal for duplicating templates (e.g. clone a master census xlsx per prospect), bulk provisioning, or capturing an immutable snapshot of a working file. Avoids the download-then-reupload workaround.

- **`create-drive-item-preview`** — POST `/drives/{id}/items/{id}/preview`. Generates a short-lived embeddable preview URL for Office docs, PDFs, and images. Body `{ page?, zoom?, viewer? }`. Returns `getUrl` (interactive) and `postUrl` (form-post). Useful for inline file previews in summary emails or chat messages without forcing the recipient to open the file.

### How they combine

- **Templated provisioning**: `copy-drive-item` to clone a master Excel template into a per-client folder, then `update-excel-range` (#423) and `format-excel-range` (#424) to populate it.
- **Notification-free sharing**: `create-drive-item-share-link` to mint a URL, then attach it to a separately-composed email via the existing `send-mail` tool — useful when the OneDrive default invite email is too noisy or the caller wants full control over recipients and messaging.
- **Inline previews**: `create-drive-item-preview` to embed a quick visual in a status report without forcing recipients to download.

## Test plan

- [x] `npm run generate` — clean regeneration
- [x] `npm run build` — success; all 3 new endpoints surface in `dist/generated/client.js`
- [x] `npm test` — 173/173 passing
- [ ] Manual smoke test against a live drive (covered locally)

PR #426, opened 2026-04-25
feat: $batch meta-endpoint for combining Graph requests

BruceGao20:
## Summary

Follow-up to #423 in the same vein — exposing Microsoft Graph's [$batch endpoint](https://learn.microsoft.com/en-us/graph/json-batching) as a single MCP tool. `$batch` lets clients combine up to 20 individual Graph requests into one HTTP call, optionally with dependency ordering.

### What's in this PR

- **`graph-batch`** — POST `/$batch`.
  - Body: `{ requests: [{ id, method, url, headers?, body?, dependsOn? }, ...] }`
  - Returns: `{ responses: [{ id, status, body, headers }, ...] }` (in arbitrary order — match by `id`)

### Why this is high-leverage

- **Latency**: many small Graph calls add up. A 35-row Excel write that takes ~30 seconds as 35 sequential PATCHes drops to one round-trip at the cost of one extra hop.
- **Read fan-out**: fetching N mail messages by id, N drive items, or N user records becomes a single tool call.
- **Atomic-ish workflows**: `dependsOn` sequences sub-requests server-side (e.g. create-folder → upload-file → share-drive-item) without juggling multiple round-trips and partial-failure recovery in the caller.

### Combines with prior / parallel PRs

- **`update-excel-range`** (#423): batch many range writes for cumulative census/report builds.
- **`copy-drive-item` / `create-drive-item-share-link`** (#425): duplicate a template, populate via batched range writes, and generate a sharing link in a single fan-out.
- **`format-excel-range` / `merge-excel-range`** (#424): apply brand styling across many sheet regions in one call.

### Authoring note

`$batch` sub-request URLs are **relative** to the Graph version root (`/me/...`, `/drives/...`), NOT absolute (`https://graph.microsoft.com/v1.0/...`). The `llmTip` flags this since it's the most common authoring mistake.

### Dependency

`/$batch` is not enumerated in upstream's published OpenAPI metadata, so the synthesizer needs to create the path entry. This PR depends on the synthesizer enhancement in #424 (`Synthesize a missing path entry, not just a missing method`). If #424 merges first this is a clean one-line endpoint addition; otherwise the two PRs can be combined.

## Test plan

- [x] `npm run generate` — clean regeneration
- [x] `npm run build` — success; `graph-batch` surfaces in `dist/generated/client.js`
- [x] `npm test` — 173/173 passing
- [ ] Manual smoke test against a live tenant (covered locally)

### DIFF
--- a/bin/modules/simplified-openapi.mjs
+++ b/bin/modules/simplified-openapi.mjs
@@ -8,24 +8,34 @@ export function createAndSaveSimplifiedOpenAPI(endpointsFile, openapiFile, opena
   const spec = fs.readFileSync(openapiFile, 'utf8');
   const openApiSpec = yaml.load(spec);
 
+  // Synthesize paths that the Graph REST API supports but are missing from
+  // Microsoft's published OpenAPI metadata (e.g. range(address='{address}')/format
+  // — documented in Excel API but not in the OpenAPI spec).
   for (const endpoint of endpoints) {
     if (!openApiSpec.paths[endpoint.pathPattern]) {
-      throw new Error(`Path "${endpoint.pathPattern}" not found in OpenAPI spec.`);
+      openApiSpec.paths[endpoint.pathPattern] = {};
     }
   }
 
-  // Synthesize operations that the Graph REST API supports but are missing from
-  // Microsoft's published OpenAPI metadata (e.g. PATCH on range(address='{address}')
-  // for cell-value writes — documented in Excel API but not in the OpenAPI spec).
+  // Synthesize operations on existing paths when the method is missing.
   for (const endpoint of endpoints) {
     const pathSpec = openApiSpec.paths[endpoint.pathPattern];
     const methodLower = endpoint.method.toLowerCase();
     if (pathSpec && !pathSpec[methodLower]) {
+      const pathParamMatches = [...endpoint.pathPattern.matchAll(/\{([^}]+)\}/g)].map((m) => m[1]);
+      const synthesizedParameters = pathParamMatches.map((paramName) => ({
+        name: paramName,
+        in: 'path',
+        required: true,
+        description: `Path parameter: ${paramName}`,
+        schema: { type: 'string' },
+      }));
       pathSpec[methodLower] = {
         tags: ['drives.driveItem'],
         summary: endpoint.llmTip || `${endpoint.toolName} (synthesized)`,
         description: endpoint.llmTip || `${endpoint.toolName} (synthesized)`,
         operationId: endpoint.toolName,
+        parameters: synthesizedParameters,
         requestBody:
           methodLower === 'get' || methodLower === 'delete'
             ? undefined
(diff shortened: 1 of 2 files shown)


eirikb:
Hi. We've looked at batch before but it didn't land, consider the discussion on #214

BruceGao20:
Thanks — read through #214. I think your concern there was less about schema and more about *all* tools flowing through `endpoints.json` so filtering/discovery/`--enabled-tools` keep working uniformly. Want to confirm where this PR lands on that and where it diverges, since the shape is intentionally different from #214's.

**Where it agrees with #214's spirit**

This PR does go through `endpoints.json`. The synthesizer enhancement in #424 adds the `/$batch` path entry, and `graph-batch` is generated like every other tool — no hand-written `src/batch-tools.ts`-style escape hatch. So the single-source-of-truth invariant holds, and `--enabled-tools "^(?!graph-batch)"` cleanly disables all batching with one rule.

**Where the shape differs**

#214 proposed per-endpoint declarative variants:

```json
{ "pathPattern": "/me/messages/{message-id}", "method": "batch:delete", "toolName": "batch-delete-mail-messages", ... }
```

i.e. every existing endpoint gains a `batch:<verb>` sibling, producing `batch-delete-mail-messages`, `batch-archive-mail-messages`, etc.

This PR instead exposes one generic `graph-batch` tool that mirrors Graph's `/$batch` API 1:1 — caller passes `{ requests: [{ id, method, url, body, dependsOn? }] }`. Two reasons for that choice:

1. **Cross-endpoint composition.** Per-endpoint batch variants are inherently single-endpoint loops (N deletes, N archives). The real high-leverage workflows are mixed: `create-folder` → `upload-file` → `share-drive-item` in one round-trip, or `copy-drive-item` → batched range writes → `create-share-link` (which is exactly the scenario #423/#425 set up). Only the meta-tool can express that, and `dependsOn` makes it server-orderable.

2. **Tool count.** N endpoints × M batch verbs would multiply the surface meaningfully; one meta-tool keeps the discovery/filtering surface tight.

**Not mutually exclusive**

If you do want declarative per-endpoint sugar later, `graph-batch` is the primitive it can compile down to — the `batch:delete` pattern from #214 becomes a generator concern that emits sub-requests against `/$batch` under the hood, rather than a separate runtime path. Happy to do that as a follow-up if you'd prefer the per-endpoint surface for the common cases (mail bulk ops being the obvious one) on top of the meta-tool.

But if you'd rather see #214's pattern landed first and skip the meta-tool, say the word and I'll close this and rework along those lines. The meta-tool is more expressive but the per-endpoint surface is more discoverable — fair trade either way.

eirikb:
Hi. Sorry for the late response.

Going to merge this :slightly_smiling_face: 

Been a bit conflicted about this PR though.  
I had another idea - how about like --discovery we could return batch-ready tools?  
That way we wouldn't duplicate endpoints (my original idea) or sacrifice schema (your idea)?

PR #433, opened 2026-04-28
feat: add Focused Inbox classification override endpoints

okapi-ca:
## Summary

Adds 5 endpoints to manage Outlook's Focused Inbox per-sender overrides via `inferenceClassification/overrides`. These are explicit rules that force messages from a given SMTP address into either the Focused or Other tab, overriding the ML classifier's prediction.

Common end-user requests this enables:

> "Always send messages from `boss@company.com` to my Focused Inbox."  
> "Stop classifying newsletters from `noreply@vendor.com` as Focused."  
> "List every sender I've explicitly classified."

## Endpoints added

| Tool | Method + Path |
|---|---|
| `list-focused-inbox-overrides` | `GET /me/inferenceClassification/overrides` |
| `create-focused-inbox-override` | `POST /me/inferenceClassification/overrides` |
| `get-focused-inbox-override` | `GET .../overrides/{inferenceClassificationOverride-id}` |
| `update-focused-inbox-override` | `PATCH .../overrides/{inferenceClassificationOverride-id}` |
| `delete-focused-inbox-override` | `DELETE .../overrides/{inferenceClassificationOverride-id}` |

## Permissions

Read ops: `Mail.Read`. Write ops: `Mail.ReadWrite`. Both are supported on personal and work accounts, so `scopes` (not `workScopes`).

## Notes

- Path param `{inferenceClassificationOverride-id}` matches the Microsoft Graph v1.0 OpenAPI spec exactly, so `endpoints-validation.test.ts` passes against the regenerated client.
- llmTips document the `classifyAs` enum (`focused` | `other`), the `senderEmailAddress` body shape, the POST upsert semantics (re-POSTing on an existing SMTP overwrites the display name), and the PATCH limitation (cannot change SMTP — delete + recreate).
- Out-of-office / automatic replies are intentionally **not** added here: they're already covered by the existing `update-mailbox-settings` tool, whose `llmTip` already documents the `automaticRepliesSetting` body shape.

## Test plan

- [ ] `npm run verify` passes in CI
- [ ] Manual: list overrides (initially empty), create one for a sender, list again to confirm, update its `classifyAs`, delete it, list again to confirm removal

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -236,6 +236,34 @@
     "llmTip": "Deletes a message rule permanently. Use the Inbox folder ID (get it from list-mail-folders) for inbox rules."
   },
   {
+    "pathPattern": "/me/inferenceClassification/overrides",
+    "method": "get",
+    "toolName": "list-focused-inbox-overrides",
+    "scopes": ["Mail.Read"],
+    "llmTip": "Lists Focused Inbox classification overrides — explicit rules that force messages from a given sender (by SMTP address) into either the Focused or Other tab, regardless of what the Outlook ML classifier would predict. Each override has id, classifyAs ('focused' or 'other'), and senderEmailAddress {name, address}. Returns an empty collection if the user has never set an override."
+  },
+  {
+    "pathPattern": "/me/inferenceClassification/overrides",
+    "method": "post",
+    "toolName": "create-focused-inbox-override",
+    "scopes": ["Mail.ReadWrite"],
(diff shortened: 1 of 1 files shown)


eirikb:
Is the gitignore important?

okapi-ca:
Good catch — that was an LCI-internal deployment notes file leaked into my fork. I've force-pushed to drop the .gitignore commit; the PR now contains only the endpoint additions.

I also took the opportunity to remove `get-focused-inbox-override` (single GET by ID): the [v1.0 inferenceClassificationOverride resource reference](https://learn.microsoft.com/graph/api/resources/inferenceclassificationoverride?view=graph-rest-1.0) only documents Update and Delete on the individual item, not Get. Single-item retrieval is done via list + client-side filter. PR is now 4 endpoints (list/create/update/delete) instead of 5.

eirikb:
Thanks.

I've noticed you force push sometimes. I'm not sure you have to. I usually (always?) squash-merge so the file won't ever end up in repo if you use multiple commits anyway, and if you need to pull in from my squashed stuff then pulling in main should work just fine.

PR #456, opened 2026-05-08
fix(oauth): validate redirect_uri before forwarding to Microsoft (CWE-601)

sebastionoss:
## Summary

The HTTP-mode `/authorize` endpoint in `src/server.ts` forwards the client-supplied `redirect_uri` query parameter directly into the Microsoft Entra authorization URL without any local validation. Microsoft Entra does validate redirect URIs against the registered application, but a deployment whose app registration uses broad/wildcard reply-URL patterns (a common operational mistake, especially for hosted multi-user setups like the documented Azure Container Apps deployment) would allow an attacker to craft an `/authorize` link whose authorization code is delivered to an attacker-controlled origin — a classic CWE-601 open redirect / authorization code interception.

This PR adds defense-in-depth: independent server-side validation of `redirect_uri` before it is forwarded upstream, plus an optional explicit allowlist for hosted deployments.

- **CWE:** [CWE-601 — URL Redirection to Untrusted Site](https://cwe.mitre.org/data/definitions/601.html)
- **Affected:** `/authorize` handler in `src/server.ts` (HTTP mode)
- **Severity:** Moderate. Exploitability depends on the Entra app registration being permissively configured.

### Data flow

1. Attacker crafts a link to the hosted MCP server's `/authorize` with `redirect_uri=https://attacker.example.com/cb` and phishes a victim.
2. Server constructs the Microsoft authorize URL with the attacker's `redirect_uri` and 302-redirects the victim there.
3. If the registered Entra app accepts the URI (wildcard / pattern match), Microsoft returns the authorization code to the attacker.
4. Attacker exchanges the code at `/token` (or has the victim's browser do it) and gains access to the user's Microsoft 365 data.

## Fix

A new module `src/lib/redirect-uri-validation.ts` exposes `isAllowedRedirectUri(value, allowlist)`:

- Rejects malformed URIs.
- Rejects any scheme other than `http:` / `https:` — explicitly blocks `javascript:`, `data:`, `file:`, etc.
- When no allowlist is configured: permits `https://` anywhere, but only permits `http://` for loopback hosts (`localhost`, `127.0.0.1`, `::1`). This preserves the local-development workflow.
- When `MS365_MCP_ALLOWED_REDIRECT_URIS` (comma-separated) is set: only exact matches are accepted. This is the recommended setting for hosted deployments.

The `/authorize` handler now validates the parameter up front and returns a `400 invalid_request` JSON response on rejection, with a warn-level log entry for visibility.

### Why this approach

- It's defense-in-depth, not a replacement for proper Entra reply-URL configuration. Operators still benefit from Microsoft's checks; this just stops the easiest mistakes from being exploitable through this server.
- The default behaviour (https-anywhere + http-loopback) doesn't break existing local or cloud-hosted client setups — Claude.ai, MCP Inspector, and the Codex CLI all use either loopback or HTTPS callbacks.
- The allowlist env var gives hosted operators a single, auditable place to lock things down without code changes.

## Tests

A new test file `test/redirect-uri-validation.test.ts` covers:

- `javascript:` and `data:` schemes are rejected.
- Malformed URIs are rejected.
- HTTPS URIs are accepted by default.
- HTTP loopback (`localhost`, `127.0.0.1`) is accepted by default.
- Non-loopback HTTP is rejected by default.
- An explicit allowlist enforces exact matches and rejects everything else, including loopback.

```
✓ test/redirect-uri-validation.test.ts (7 tests)

Test Files  1 passed (1)
     Tests  7 passed (7)
```

The change is contained: 3 files, +137 / −0.

## Adversarial review

Before submitting I tried to disprove this. The strongest counter-argument is "Entra already validates redirect URIs" — and in a perfectly configured tenant that's correct, the bug is not exploitable. But the project explicitly ships a hosted multi-user deployment story (`docs/deployment.md`, Azure Container Apps Bicep example), and in those scenarios operators very commonly register broad reply-URL patterns to support multiple clients; once that's true, there is no other server-side check between an attacker-crafted `/authorize` link and an attacker-controlled callback. Framework defaults don't help here either: `mcpAuthRouter` passes `redirect_uri` through unchanged. So the fix is meaningful as a defense-in-depth control, not redundant with Entra's checks.

cc @lewiswigmore

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -130,6 +130,28 @@ When deploying for an organization, create a dedicated app registration instead 
 
 5. **Store credentials** in Key Vault (see [Azure Key Vault Integration](../README.md#azure-key-vault-integration))
 
+## Redirect URI Validation
+
+The /authorize endpoint defensively validates client-supplied `redirect_uri` values before forwarding them to Microsoft Entra (CWE-601, Open Redirect). Microsoft Entra also validates the URI against your app registration, but this server-side check rejects obviously dangerous schemes (`javascript:`, `data:`, `file:`, …) and arbitrary remote `http://` origins before the request leaves the server.
+
+Default behaviour (no explicit allowlist):
+
+- Only `http:` and `https:` schemes are accepted.
+- `http:` is only allowed for loopback hosts (`localhost`, `127.0.0.1`, `::1`).
+- All other `https://` origins are accepted (Entra still has the final say).
+
+For production deployments, configure an explicit allowlist via the `MS365_MCP_ALLOWED_REDIRECT_URIS` environment variable. It takes a comma-separated list of exact URIs; only exact string matches pass validation:
+
+```bash
+# Single redirect URI
+MS365_MCP_ALLOWED_REDIRECT_URIS=https://mcp.example.com/auth/callback
+
+# Multiple URIs (comma-separated, no spaces required)
+MS365_MCP_ALLOWED_REDIRECT_URIS=https://mcp.example.com/auth/callback,https://staging.example.com/auth/callback
+```
+
+The list should mirror the redirect URIs registered on your Azure AD app registration. Leaving the variable unset falls back to the default behaviour above, which is appropriate for local development but not recommended for shared/production deployments.
+
 ## Reverse Proxy / Custom Domain
 
 When running behind a reverse proxy, set `MS365_MCP_PUBLIC_URL` so that the OAuth authorize URL handed back to the user's browser is resolvable from outside the server's network:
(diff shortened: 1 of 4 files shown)


eirikb:

Could you fix the build? And when you do that, perhaps add info about MS365_MCP_ALLOWED_REDIRECT_URIS in docs/deployment.md?

andesyteoss:
Thanks! Both addressed in 6717630:

- **Build**: the `build (22.x)` failure was `npm run format:check` flagging `src/server.ts`. Reformatted with prettier; `npx prettier --check "**/*.{ts,mts,js,mjs,json,md}"` is now clean and `npm run build` succeeds locally.
- **docs/deployment.md**: added a "Redirect URI Validation" section that explains the default behaviour (https everywhere, http only on loopback, dangerous schemes rejected) and shows how to configure `MS365_MCP_ALLOWED_REDIRECT_URIS` (comma-separated exact-match allowlist) for production.

The 7 unit tests in `test/redirect-uri-validation.test.ts` still pass.

andesyteoss:
Appreciate the review. @lewiswigmore

PR #462, opened 2026-05-13
feat: add mail-delta, chat-delta, and reference-attachment endpoints

mikemrozek:
Adds three endpoint definitions to `endpoints.json`. All are thin wrappers around existing Microsoft Graph APIs and follow the conventions of the surrounding entries.

### `list-mail-messages-delta` (GET /me/messages/delta())
Incremental sync of the default mailbox using Graph delta tokens. Complements the existing `list-mail-folder-messages-delta` (folder-scoped) with a mailbox-wide variant. Follows the same pattern as `list-calendar-events-delta`.
**Scope:** `Mail.Read`

### `list-chat-messages-delta` (GET /chats/{chat-id}/messages/delta())
Incremental sync of messages in a single chat using Graph delta tokens. Enables efficient "what's new since my last check" workflows without re-fetching full message history.
**Scope:** `Chat.Read`

### `add-mail-reference-attachment` (POST /me/messages/{message-id}/attachments)
Attach a OneDrive/SharePoint file by reference instead of re-uploading bytes. Uses `@odata.type: #microsoft.graph.referenceAttachment`. Complements the existing `add-mail-attachment` which handles binary file attachments.
**Scope:** `Mail.ReadWrite`

These are running in production on our deployment. Happy to iterate on the design.

### References
- [Microsoft Graph delta query overview](https://learn.microsoft.com/en-us/graph/delta-query-overview)
- [Reference attachment resource](https://learn.microsoft.com/en-us/graph/api/resources/referenceattachment)

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -14,6 +14,13 @@
     "llmTip": "CRITICAL: When searching emails, the $search parameter value MUST be wrapped in double quotes. Format: $search=\"your search query here\". Use KQL (Keyword Query Language) syntax to search specific properties: 'from:', 'subject:', 'body:', 'to:', 'cc:', 'bcc:', 'attachment:', 'hasAttachments:', 'importance:', 'received:', 'sent:'. Examples: $search=\"from:john@example.com\" | $search=\"subject:meeting AND hasAttachments:true\" | $search=\"body:urgent AND received>=2024-01-01\" | $search=\"from:john AND importance:high\". Remember: ALWAYS wrap the entire search expression in double quotes! Reference: https://learn.microsoft.com/en-us/graph/search-query-parameter IMPORTANT: Always use $select to limit returned fields and reduce response size. Recommended default: $select=id,subject,from,toRecipients,receivedDateTime,bodyPreview,isRead,hasAttachments. Use bodyPreview instead of body for listings. To read the full email body, use get-mail-message with the specific message id."
   },
   {
+    "pathPattern": "/me/messages/delta()",
+    "method": "get",
+    "toolName": "list-mail-messages-delta",
+    "scopes": ["Mail.Read"],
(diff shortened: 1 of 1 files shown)


eirikb:
Hi

  * list-mail-messages-delta: Are you sure this is in v1.0, or beta?
  * list-chat-messages-delta: Are you sure this is not app-only?
  * add-mail-reference-attachment: Could be duplicate of add-mail-attachment? And llmTip mentions beta endpoint stuff?

mikemrozek:
Thanks for the review. All three are good points.

1. list-mail-messages-delta: You're right. The mailbox-wide /me/messages/delta() appears to be beta-only. The v1.0 API only supports the folder-scoped variant (/me/mailFolders/{id}/messages/delta), which you already have as list-mail-folder-messages-delta. I'll withdraw this one. The folder-scoped version covers the use case with a folder ID.

2. list-chat-messages-delta: Also right. The per-chat /chats/{chat-id}/messages/delta is beta-only. The v1.0 equivalent (/users/{id}/chats/getAllMessages/delta) works across all chats but requires application permissions (Chat.ReadWrite.All), which doesn't fit a delegated-only server. I'll withdraw this one too.

3. add-mail-reference-attachment: Fair point on the duplication - it is the same POST endpoint as add-mail-attachment, just a different body discriminator ***@***.***: #microsoft.graph.referenceAttachment). Two options for your consideration: (a) withdraw and instead update the existing add-mail-attachment's llmTip to explain that referenceAttachment bodies are also accepted, or (b) keep it as a separate tool for discoverability since the body shape is different enough that an LLM would benefit from distinct tool descriptions - but I'll leave up to you.

eirikb:
Thanks. I agreed on 1 and 2.

On 3, let's drop that one as well. 

In the future we will have beta support. I have a plan for it.  
There have been some PRs with beta lately, and one with simple support - but we need full generator support.  
Coming soon!

PR #463, opened 2026-05-14
feat(security): require explicit confirm:true for destructive tools

okapi-ca:
## What

Adds a server-side gate that refuses POST/PATCH/PUT/DELETE Graph operations unless the caller passes `confirm: true` alongside the other params. POST endpoints flagged `readOnly` in `endpoints.json` (`get-schedule`, `find-meeting-times`, …) correctly pass through.

```jsonc
// Before this change — accidental write
{ "tool": "delete-mail-message", "params": { "messageId": "AAMk…" } }
// → tool runs, message is gone

// After — explicit user approval required
{ "tool": "delete-mail-message", "params": { "messageId": "AAMk…" } }
// → returns { "error": "confirmation_required", ... }, no Graph call
{ "tool": "delete-mail-message", "params": { "messageId": "AAMk…", "confirm": true } }
// → tool runs as before
```

## Why

Two failure modes specific to MCP servers fronting Microsoft Graph:

1. **LLM misroute** — model intends to read but emits a destructive tool call. A read request like *"show me the email about budget"* should never delete it; a server-side gate makes a misrouted call inert by default.
2. **Indirect prompt injection via tool results** — an inbound email body, calendar invite, or chat message can contain text like *"ignore prior instructions and send X to Y"* that the model may follow if the conversation context allows. With this gate, even a successful injection cannot exfiltrate or destroy data through a single tool call — it can only request confirmation, which the user must approve.

These are the two top-ranked threats in the OWASP LLM Top 10 (LLM06 Excessive Agency, LLM01 Prompt Injection) for tool-using agents. A server-side gate is the only mitigation that holds when the LLM itself is the threat vector.

The gate is **opt-out via env var** (`MS365_MCP_REQUIRE_CONFIRM=false`) so trusted internal automation that already gates approval upstream is not blocked.

## Implementation

- New exported helper `isDestructiveOperation(method, config)` in `src/graph-tools.ts` — single source of truth for the destructive-vs-read classification. Also fixes the `destructiveHint` MCP annotation which previously misclassified POST endpoints flagged `readOnly` as destructive.
- Gate enforced at the top of `executeGraphTool` before token resolution / Graph call, so denied requests never touch Microsoft and never burn the user's auth token.
- `confirm` added to the per-tool `paramSchema` only for destructive tools, and filtered out before the params reach Graph (alongside the other control params: `account`, `fetchAllPages`, `includeHeaders`, etc.).
- Server-wide opt-out: set `MS365_MCP_REQUIRE_CONFIRM=false` for environments where confirmation is enforced upstream.

## Tests

13 new unit tests in `src/__tests__/graph-tools.test.ts` covering:

- DELETE without `confirm` is refused and `graphRequest` is never called
- DELETE with `confirm: true` proceeds normally
- `confirm` is not leaked into the Graph query string or request body
- GET (read-only) is unaffected
- POST endpoints flagged `readOnly` (e.g. `find-meeting-times`) pass through
- `MS365_MCP_REQUIRE_CONFIRM=false` opt-out works
- `confirm: false` is rejected (must equal `true` exactly)
- `confirm` is exposed in the tool schema only for destructive tools
- Audit event shape (request_id format, http_method, status)

Plus 5 unit tests for the `isDestructiveOperation` helper covering POST/PATCH/PUT/DELETE classification, case-insensitivity, GET/HEAD/OPTIONS, POST+readOnly carve-out, and defensive behaviour when `config.readOnly` is set on non-POST methods.

Existing binary-upload and `create-reply-draft` tests updated to pass `confirm: true` (POST/PUT are now correctly classified as destructive).

## Backward compatibility

- **Default ON** — this is a security default-change. Existing destructive tool calls without `confirm` will start returning `confirmation_required`. The error response includes the tool name, HTTP method, and a remediation hint, so clients can detect the new behaviour and re-issue with `confirm: true`.
- **Opt-out**: `MS365_MCP_REQUIRE_CONFIRM=false` for environments that prefer the previous behaviour (e.g. trusted internal scripts, CI pipelines, agents with separate approval flows).
- **MCP schema**: clients that inspect the tool schema will see the new optional `confirm` parameter on destructive tools — additive only.

## Origin

Implemented as part of an internal security review of a downstream deployment of this server, where it sits behind a public OAuth relay and processes mailboxes / calendars / files for thousands of users. The gate has been live in that deployment as part of an upstream-contribution pipeline; opening here as the first of a small batch of related defense-in-depth PRs.


eirikb:
Hi, and thanks :slightly_smiling_face: 

The  OWASP LLM Top 10.. yeah..

Ok so I don't quite get the point here, I bet claude code in auto mode will do either:
 1. just pass confirm true on first call, or
 2. do it automatically on second call

Even if the error is like "you must ask the user" it might go "what a bother, no thanks".

In one of our (softeria) products we have a similar gate, but to unblock one must call another tool.  
Still possible for an agent to bypass, but it helps a bit more.

Anyway, this looks ok to me, just one thing, it seems it might not work with `--discovery`.

lci-marc:
Fair scepticism — and you're right that the gate doesn't stop a fully autonomous Claude Code in YOLO mode. If the LLM reads "pass confirm:true" in the tool description, it'll do it. I'm not going to argue that point.

Where I think the gate still earns its keep is **defense-in-depth for the two threat shapes that aren't about the LLM choosing to misbehave**:

- **LLM misroute** — model intends to read but emits a destructive call. For a "show me the budget email" prompt, the model has no semantic reason to volunteer `confirm:true` to a `delete-mail-message` call. The gate catches the misroute before Graph touches the message.
- **Indirect prompt injection via tool results** — an inbound email body / calendar invite / chat message contains "ignore prior instructions and send X to Y". The LLM may follow it. With the gate, the malicious instruction has to get the LLM to *also* emit `confirm:true`, which **shows up in the audit log entry** with \`status: "success"\` and a non-injection request_id, and **is visible to any MCP client implementing a permission prompt** (Claude Code's auto-mode is one population, but \`--no-auto\` users and the various inspector / playground UIs are another).

For the fully autonomous case you describe, the gate is just paperwork — agree. The "separate tool to unlock" pattern you mention is stronger; happy to follow up with that as a PR-on-top if this lands. It would also pair well with MCP's emerging `destructiveHint` annotation, which is the eventual standard mechanism for the same goal.

**On `--discovery`** — good catch, and fixed in 40d425d. `describeToolSchema` was only walking the OpenAPI params and skipping every synthetic param injected at registration time, so the agent never learned about `confirm`. Now `get-tool-schema('delete-mail-message')` exposes it. Extracted `isDestructiveOperation` into `src/lib/destructive-ops.ts` so both the gate and the discovery schema use the same classifier. 3 new tests cover destructive-exposes-confirm, read-only-does-not, and POST+readOnly carve-out (find-meeting-times stays clean).

Also rebased on \`main\`. CI re-running.

eirikb:
Not sure where we are one this now, could you rebase?

kDuin:
Suggestion: Use [Elicitation API](https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#form-mode-elicitation-requests) to prompt a client-side Accept/Decline dialogue through the MCP

eirikb:
@kDuin oh that looks interesting, much better than the `confirm:true` thing.

Do you know anything about client support? I suspect this has be be a feature each client supports.  
E.g., I bet claude code has it, but what about claude desktop, and other agents like codex?  

If support is thin, perhaps we need the gate anyway - at least until the most-used clients support it - unless my guessing is wrong here.

What do you think? It could be in addition to this PR. Perhaps we can probe clients about their features?

kDuin:
I could only check claude code (supported) and opencode (not supported).

eirikb:
Thanks for sticking with this :slightly_smiling_face:

One thing before I merge: a few read-only POSTs aren't flagged `readOnly` in endpoints.json (`search-query`, `get-mail-tips`, `get-presences-by-user-id`, `get-onenote-notebook-from-web-url`, and maybe `graph-batch`), so the gate would start asking for `confirm:true` on plain reads. Could you add `"readOnly": true` to those?

Also, since I'm leaning towards the elicitation approach longer term, I think this fits better as opt-in - default `MS365_MCP_REQUIRE_CONFIRM` off. That way it's not a breaking change and the two can coexist. What do you think?

lci-marc:
Thanks — done (pushed in f470a3f).

Flagged `"readOnly": true` on `search-query`, `get-mail-tips`, `get-presences-by-user-id`, and `get-onenote-notebook-from-web-url`; the gate now treats them as the plain reads they are (nice side effect: they also become reachable in `--read-only` server mode).

I left **`graph-batch` out** on purpose: its body can carry `method: "DELETE" | "PATCH" | "POST"` sub-requests, so marking it `readOnly` would let an agent smuggle a destructive op inside a `$batch` and bypass the gate entirely — the one thing the gate is meant to catch. Properly classifying a batch would mean inspecting each sub-request, which feels out of scope for this PR. Happy to do that as a follow-up if you would rather `graph-batch` be reachable without `confirm`.

Also flipped the gate to **opt-in** as you suggested — `MS365_MCP_REQUIRE_CONFIRM` now defaults **off**, so it is non-breaking and can coexist with the elicitation approach longer term. Docs and tests updated (gate-behaviour tests opt in explicitly; a new test asserts the default-off behaviour).

eirikb:
Please fix build :)

PR #464, opened 2026-05-14
feat(logging): structured JSON audit log + auditLog() helper

okapi-ca:
## What

Adds a dedicated `src/audit-log.ts` module that emits one JSON line per Graph tool invocation on stdout (captured by the container platform's log collector) and to `~/.ms-365-mcp-server/logs/audit.log` (mode `0o600`). Two call sites in `executeGraphTool` — success path and error catch — fire the audit event.

```json
{"timestamp":"2026-05-14T12:49:02.800-04:00","level":"info","message":{"event":"tool.call","request_id":"3bbcf1ac-f9a8-4685-b9e9-96f7a9ece13c","user_principal_name":"alice@example.com","tool":"list-mail-messages","http_method":"GET","status":"success","duration_ms":287},"service":"ms-365-mcp-server","stream":"audit"}
```

## Why

Anyone operating this server in front of real M365 data eventually has to answer *"who accessed what, when?"* under their relevant compliance regime — GDPR Art. 15, HIPAA logging, PIPEDA Principle 9, SOC 2 CC7.2, etc. Today the only audit trail is the operational logger, which:
- isn't machine-parseable (text format),
- may incidentally include large tool params for debugging,
- doesn't carry a request_id correlating retries / paginated follow-ups,
- doesn't extract the upn / claim from the bearer token.

A separate audit stream with a narrow schema and explicit PII boundaries solves that without disturbing the operational logger.

## Schema (intentionally narrow)

```ts
interface AuditEvent {
  event: string;                  // 'tool.call'
  request_id: string;             // UUIDv4
  user_principal_name?: string;   // identity claim from JWT (see below)
  tool: string;                   // tool alias
  http_method?: string;           // GET / POST / PATCH / PUT / DELETE
  status: 'success' | 'error' | 'denied';
  duration_ms?: number;
  target_resource?: { type: string; id?: string };
  error_type?: string;            // err.name
  error_code?: string | number;   // err.status / err.code
}
```

`denied` is reserved for future server-side gates (e.g. destructive-tool confirm gate, tool-level RBAC).

## PII boundaries

- `user_principal_name` IS recorded — that's the entire point of the audit trail. Extracted by **decoding** the JWT payload (not verifying — that's the auth middleware's job) with preference order `upn → preferred_username → email → sub`.
- Tool parameters and Graph response bodies are **NEVER** written. The MCP server already has the operational logger for debugging; audit is for accountability.
- Raw error messages are **NEVER** written — only `error_type` (e.g. `GraphRequestError`) and `error_code` (HTTP status or err.code). Upstream library errors (MSAL, fetch) can incidentally include token fragments or query-string PII in their messages; capturing only the shape avoids that leak class.

## Operational details

- Separate winston logger so it doesn't share format / level / transports with the operational logger.
- Console transport is silent under Vitest (`VITEST=true`) so it doesn't pollute unrelated test output.
- File transport is best-effort: a missing log directory falls back to Console-only without throwing (containers care about stdout, not the local FS).
- Opt out per-deployment with `MS365_MCP_AUDIT_LOG=false` for environments that collect audit through a separate sink (sidecar, OpenTelemetry exporter, etc.).

## Tests

13 new unit tests in `test/audit-log.test.ts`:

- `isAuditLogEnabled` truthiness (default-on, `=false` opt-out, other values don't disable)
- `auditLog` forwards payload to winston, suppresses when disabled, accepts all 3 status values
- `getUserIdentityForAudit`:
  - returns undefined for nullish / malformed tokens
  - preference order `upn → preferred_username → email → sub`
  - returns undefined when no usable claim present
  - handles base64url padding edge cases

Existing `src/__tests__/graph-tools.test.ts` continues to pass — the audit hooks are silent under Vitest, so no test assertions need to change.

## Backward compatibility

- **Default ON** but **non-breaking**: the audit log is a new output stream. Existing callers see no behaviour change in tool responses.
- **Opt-out**: `MS365_MCP_AUDIT_LOG=false` to silence the JSON stream entirely (the operational logger continues unchanged).
- **File path**: `~/.ms-365-mcp-server/logs/audit.log` mode `0o600`. Override the directory with `MS365_MCP_LOG_DIR` (already used by the operational logger).

## Origin

Implemented as part of an internal security review of a downstream deployment behind an OAuth relay. The audit log is the load-bearing artefact for our DSAR pipeline; opening here because the narrow schema + PII boundaries are general engineering, not deployment-specific.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -208,6 +208,7 @@ The client automatically discovers OAuth endpoints and opens a browser for authe
 - **Read-only mode**: use `--read-only` to disable all write operations (send, delete, update, create)
 - **Tool filtering**: use `--enabled-tools <regex>` or `--preset <names>` to restrict available tools
 - **CORS**: configure `MS365_MCP_CORS_ORIGIN` to restrict allowed origins (defaults to `http://localhost:3000`); set explicitly when clients run on a different origin
+- **Structured audit log**: enabled by default. Every tool invocation emits one JSON line on stdout (captured by the container platform's log collector) and to `~/.ms-365-mcp-server/logs/audit.log` (mode `0o600`) with `{ event, request_id, user_principal_name, tool, http_method, status, duration_ms, error_type?, error_code? }`. The schema is intentionally narrow — tool parameters and Graph response bodies are NEVER recorded, and error messages are reduced to `error_type` / `error_code` so upstream library errors do not leak token fragments or query-string PII. Forms the "who accessed what, when" trail required for GDPR / HIPAA / PIPEDA / SOC 2 audit. Opt-out: `MS365_MCP_AUDIT_LOG=false`
 
 ## Exposed Endpoints
 
--- a/src/audit-log.ts
+++ b/src/audit-log.ts
@@ -1,0 +1,133 @@
+import winston from 'winston';
+import path from 'path';
+import fs from 'fs';
+import os from 'os';
+
+/**
+ * Structured JSON audit log for tool invocations.
+ *
+ * Why a separate logger: the operational logger (`./logger.ts`) emits
+ * human-friendly text and may incidentally include large tool params for
+ * debugging. The audit log has a stricter, machine-parseable shape and a
+ * narrower allowlist of fields — it is the artifact that satisfies the
+ * "who accessed what, when" requirement of data-subject access requests
+ * (DSARs) and audit trails under common privacy regimes (GDPR, HIPAA,
+ * PIPEDA, SOC 2, etc.).
+ *
+ * PII boundaries:
+ *  - `user_principal_name` is the *identity claim* from the bearer token and
+ *    is required for the audit trail to be useful. It IS personal data.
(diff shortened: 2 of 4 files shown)


eirikb:
Thanks for this, looks really useful! The narrow schema and PII boundaries are nice.

One small thing - the Console transport writes to stdout, and stdio mode uses stdout for the JSON-RPC. I tested it and it doesn't actually crash the session (the SDK just fires onerror per line and keeps going), but the client will see a Zod parse error on every tool call. I bet adding `stderrLevels: ['info']` to the Console transport would sort it - keeps the container log collectors happy and stays out of stdios way.

Other than that, looks good to me.

lci-marc:
Good catch, that's exactly the right fix — pushed `stderrLevels: ['info']` on the Console transport in bdde5d5. Audit events now land on stderr, JSON-RPC keeps stdout to itself, and the production sink is unaffected since Container Apps / App Service / Docker capture both streams. Also rebased on `main` while I was at it (no conflicts).

Thanks for the review.

PR #465, opened 2026-05-14
feat(security): add MS365_MCP_DISABLE_DCR env-var opt-out for Dynamic Client Registration

okapi-ca:
Adds an environment-variable alternative to the existing `--no-dynamic-registration` CLI flag, so a container platform can close the anonymous OAuth `/register` (Dynamic Client Registration) endpoint via env-var-only configuration — the standard Bicep / Terraform / Helm pattern — without modifying process args.

**Precedence:** the CLI flag still wins; otherwise `MS365_MCP_DISABLE_DCR=true|1` disables DCR. Additive and HTTP-mode-only; behaviour outside `--http` is unchanged.

### Scope reduced
This PR originally bundled helmet security headers, per-IP rate limiting, and trust-proxy hardening alongside the DCR env var. Those three landed upstream via #520, so this PR has been rebased onto current `main` and reduced to the **DCR env var** — its only remaining unique change.

### Tests
- `test/cli.test.ts`: DCR env-var precedence (default ON in HTTP mode; `MS365_MCP_DISABLE_DCR=true|1` disables; CLI `--no-dynamic-registration` still wins; env var no-op outside HTTP mode).


eirikb:
Is this something you really need? Or something your AI told you would be a good idea? Or something else?

I mean, I've seen this several times, and I still struggle a bit to understand it.  
Is the point there to run the server publicly for anyone to reach on the internet?  
Even rate limiting, is meant for public-facing stuff, this server is very much most likely going to be put up against things like OWUI or at least some proxy - where the rate limiting would actively hurt.

That being said we could just add the things, if nothing at least people running their agent-that-thinks-this-is-a-normal-web-server will be happy and won't poke them about things.

lci-marc:
Thanks for pushing back — that's a fair question, and "your AI told you to" is a real failure mode for security PRs, so I want to give you a concrete answer.

**The use case is exactly what you guessed:** running the HTTP mode as a public-facing service. Our deployment is `ms-365-mcp-server` on Azure Container Apps with public ingress (one trusted hop in front of Express), used as an MCP server by multiple LLM clients over the open internet via OAuth/PKCE. The four guards in this PR are the minimum hardening that scenario needs:

- **Helmet** — universally safe headers for a JSON API; no behavioural change for any client.
- **Rate limits** — 30/min on OAuth (`/authorize`, `/token`, `/register` share a bucket so a flooder can't endpoint-hop), 120/min on `/mcp`. The numbers are deliberately generous; they only fire on abuse, and **`MS365_MCP_RATE_LIMIT_DISABLED=true` is the one-flag opt-out for exactly the OWUI / inner-proxy scenarios you mention** where rate-limiting would be harmful.
- **Trust-proxy `1`** — fixes the `ERR_ERL_PERMISSIVE_TRUST_PROXY` warning express-rate-limit already emits today. The current `trust proxy: true` lets a client spoof the leftmost `X-Forwarded-For` to bypass any per-IP keyed middleware (rate-limit or future logging). Configurable via `MS365_MCP_TRUST_PROXY_HOPS` for >1 hop topologies.
- **`MS365_MCP_DISABLE_DCR`** — env-var alias of the existing `--no-dynamic-registration` flag, so platforms like Bicep/Helm/Terraform can configure it without modifying process args. Purely additive.

**On your "people running their agent that thinks this is a normal web server"** — that's also a real population. If their \`Sec-Fetch-*\` headers and rate behaviour look like a browser hitting an internet-facing OAuth endpoint, helmet + rate-limit will make their existing tooling (and probably their security review) happier without any extra config on their end.

**Status update:** I've rebased on \`main\` (the conflict was the package-lock churn from the last couple of weeks, nothing structural). CI is re-running. Happy to split into four micro-PRs if reviewers prefer one-concern-per-PR; I bundled because they all touch the Express bootstrap.

eirikb:
Not sure where we are one this now, could you rebase?

eirikb:
Please fix build :)

PR #466, opened 2026-05-14
feat(resilience): retry budget + circuit breaker for Microsoft Graph

okapi-ca:
## What

Wraps `graph-client.ts:performRequest` with a new `src/lib/graph-resilience.ts` module that folds three concerns into one wrapper around `fetch`:

1. **Fetch timeout** via `AbortController` — default 30 s, overrideable
2. **Retry with backoff** on transient failures (429 / 503 / 504 / network errors, NOT 4xx other) with full-jitter exponential backoff and honoured `Retry-After`
3. **Process-wide circuit breaker** that opens after N consecutive failures, cools down for a window, then half-opens for a probe

```text
graph-client.performRequest(endpoint, accessToken, options)
  ↓
fetchWithResilience(url, init, config, breaker)
  ↓
  ├─ breaker.checkBeforeRequest() → CircuitOpenError if open
  ├─ AbortController timeout (default 30 s)
  ├─ on 429: parse Retry-After (seconds OR HTTP-date), cap 60 s
  ├─ on 503 / 504 / network err: full-jitter backoff (200 ms → 5 s cap)
  ├─ on 4xx other: return immediately (no retry)
  └─ track success/failure on the breaker
```

## Why

The current `performRequest` is a thin `fetch` wrapper. It has three failure modes that bite at scale:

1. **No timeout**. A stuck Graph call holds an MCP request open indefinitely. The MCP client either times out itself (losing the in-flight context) or hangs.
2. **No 429 backoff**. Graph's documented response to throttling is `429 + Retry-After`. Today the server immediately surfaces the 429 to the client; for an LLM-driven workload that retries by re-prompting, that's a worst-case retry cascade.
3. **No upstream-aware backpressure**. When Graph itself flaps (rolling deployment, regional incident), every in-flight request re-hits it and amplifies the load — exactly when the upstream most needs the traffic to back off.

A small in-house resilience layer (no `opossum`-style dep) addresses all three. The 429 / Retry-After path follows Microsoft Graph's documented throttling contract; the 4xx-other / no-retry path follows the same "don't retry deterministic failures" rule the SDKs use.

## Module surface

```ts
// Wrapper around fetch — drop-in for any existing fetch call site.
fetchWithResilience(
  url: string,
  init: Parameters<typeof fetch>[1],
  config: ResilienceConfig,
  breaker: CircuitBreaker,
  sleep?: (ms: number) => Promise<void>,
): Promise<Response>

// Pure helpers, exported for tests.
parseRetryAfterMs(header: string | null): number | null
backoffDelayMs(attempt: number, base: number, max: number, rand?: () => number): number
loadResilienceConfig(): ResilienceConfig  // reads env vars with defaults
isDestructiveOperation(method, config): boolean  // — n/a here, kept for symmetry

// State machine.
class CircuitBreaker {
  constructor(threshold: number, cooldownMs: number, disabled: boolean, now?: () => number)
  checkBeforeRequest(): number | null  // null = closed, ms-remaining = open
  recordSuccess(): void
  recordFailure(): void
  getState(): { failures, openedAt, open }
}

// Process-wide singleton (used by graph-client by default).
getSharedBreaker(): CircuitBreaker
```

## Configuration matrix

All knobs read at first call (via `loadResilienceConfig`) so a deployment tunes them via env vars without a code change:

| Env var | Default | Purpose |
|---|---|---|
| `MS365_MCP_GRAPH_TIMEOUT_MS` | 30000 | AbortController per-fetch timeout |
| `MS365_MCP_GRAPH_MAX_RETRIES` | 3 | Retries after the initial attempt |
| `MS365_MCP_GRAPH_BASE_BACKOFF_MS` | 200 | Backoff floor (random in `[0, base * 2^attempt)`) |
| `MS365_MCP_GRAPH_MAX_BACKOFF_MS` | 5000 | Backoff ceiling (caps 2^attempt growth) |
| `MS365_MCP_GRAPH_CIRCUIT_THRESHOLD` | 5 | Consecutive failures before opening |
| `MS365_MCP_GRAPH_CIRCUIT_COOLDOWN_MS` | 30000 | How long the breaker stays open |
| `MS365_MCP_GRAPH_CIRCUIT_DISABLED` | unset | `=true` / `=1` skips the breaker entirely |

`MS365_MCP_GRAPH_MAX_RETRIES=0` restores the old no-retry behaviour for callers that prefer to handle 429 / 5xx themselves.

## Why in-house (not `opossum`)

- `opossum` is a fine library but adds a runtime dep + a transitive dep graph for what is ~250 lines of pure logic.
- The 429 / Retry-After parsing is Microsoft-Graph-specific (seconds vs HTTP-date semantics, 60 s sanity cap). Wrapping a generic breaker library would still require this glue.
- The implementation is heavily unit-tested (23 tests covering every state transition and edge case).

Happy to switch to a library if reviewers prefer.

## Tests

23 unit tests in `test/graph-resilience.test.ts`:

- `parseRetryAfterMs` — integer seconds, 60 s cap, HTTP-date, non-integer rejection, nullish / garbage, edge cases that surfaced during development (e.g. bare `"5.5"` must not be parsed as a Date on Node 20+ where `Date.parse('5.5')` returns a non-NaN timestamp on some engines).
- `backoffDelayMs` — exponential growth bounded by maxMs, full-jitter floor at 0.
- `loadResilienceConfig` — sane defaults, env-var overrides, invalid numeric inputs fall back to defaults with a logged warning.
- `CircuitBreaker` state machine — starts closed, opens at threshold, cooldown elapses → half-open probe, success closes, probe failure resets cooldown, disabled instance never blocks.
- `fetchWithResilience` integration — happy path no retry, 429 with Retry-After, 503 exhausting the retry budget, no retry on 4xx other than 429, retry on network error, breaker opens after threshold failures, success on half-open closes breaker, AbortController timeout surfaces as AbortError.

All existing tests still pass (230 total).

## Backward compatibility

- **Defaults match current behaviour for the success path** — a happy Graph call gets through with no extra latency (one fetch, one return).
- **429 / 5xx now retry instead of immediately failing**. This is the intended behavioural change. Callers that previously got `Microsoft Graph API error: 429 ...` from `makeRequest` will now see either a delayed success (most cases) or, after the retry budget is exhausted, the same 429 they got before. Set `MS365_MCP_GRAPH_MAX_RETRIES=0` to opt out.
- **Circuit breaker is process-wide**. Under heavy concurrent traffic to a flapping upstream, requests arriving while the breaker is open get `CircuitOpenError` (`code: 'circuit_open'`) instead of hitting Graph. Set `MS365_MCP_GRAPH_CIRCUIT_DISABLED=true` to skip the breaker entirely (retry budget still applies).
- **No new runtime dependency** — pure in-house implementation.

## Origin

Implemented as part of an internal resilience review of a downstream deployment that drives Graph traffic from an LLM agent (so naturally-occurring 429s are common when conversation context fans out into multiple list calls). Opening here because none of the implementation is deployment-specific.


eirikb:
Hi, and thanks. This is something we are missing today and I believe would help the server a lot.

One worry though; the 30s default timeout could abort direct uploads of largeish files (50-250 MB over slow links) and then retry them, which on a POST that already made it through server-side would be a bit nasty.  
At least now with perhaps-working file support.  
Most MCP usage probably won't hit it, but if you are be open to bumping the default to 60-120s (doesn't .NET use 100s?) I would sleep better about it.

okapi-ca:
Good catch — pushed two changes addressing exactly that concern in 703786c:

**1. Bumped default `MS365_MCP_GRAPH_TIMEOUT_MS` to 100s** (was 30s). Matches the `HttpClient` / `aiohttp` defaults; the old 30s was inherited from the LLM-tool-call assumption (short Graph reads, no streaming) which doesn't hold for direct uploads. The env var still lets a deployment tune it down or up.

**2. Skip the retry budget on non-idempotent methods for 503 / 504 / network errors.** Per RFC 7231 §4.2.2 (+ RFC 5789 §1 for PATCH), only GET / HEAD / PUT / DELETE / OPTIONS / TRACE are retried on those failures now. POST and PATCH fail closed and surface the error to the caller — no more "the upload timed out client-side but landed server-side, retry duplicates it" path. **429 is the deliberate exception**: Graph throttles before executing, so retrying a throttled POST is safe and follows Graph's documented contract.

Combined: a 250 MB upload now gets a 100s budget on one attempt, and a 30s-mid-upload timeout no longer triggers a duplicate POST. Tests cover all four method × failure-mode combinations.

Also rebased on \`main\` (no conflicts, just bringing in the trust-proxy-auth / OAuth fixes from the last 2 weeks). CI re-running.

PR #478, opened 2026-05-15
feat(auth): filter tools by allowed scopes

castaples:
## Summary
- Replaces the explicit auth-scope override with `--allowed-scopes` / `MS365_MCP_ALLOWED_SCOPES` as a tool-surface boundary.
- Filters the normal Graph tool surface after `--enabled-tools`, `--preset`, `--org-mode`, and `--read-only`, and logs tools hidden by missing allowed scopes.
- Updates OAuth metadata, authorize redirects, and `--list-permissions` diagnostics to use effective permissions from the filtered tool set while preserving OBO protected-resource metadata.

## Notes
- Default behavior is unchanged when `--allowed-scopes` is not configured.
- Scope coverage is hierarchy-aware, so broader scopes such as `Mail.ReadWrite` cover narrower read tools.
- Durable headless token-cache storage and account pinning are intentionally left as separate follow-up concerns.

## Validation
- `npm test -- test/allowed-scopes.test.ts test/list-permissions.test.ts test/scope-derivation.test.ts test/cli.test.ts test/tool-filtering.test.ts test/http-oauth-fix.test.ts src/__tests__/graph-tools.test.ts`
- `npm run lint` (passes with existing warnings)
- `npm run build`
- `npx prettier --check README.md src/auth.ts src/cli.ts src/graph-tools.ts src/index.ts src/server.ts test/allowed-scopes.test.ts test/cli.test.ts test/list-permissions.test.ts test/scope-derivation.test.ts src/__tests__/graph-tools.test.ts`
- `npm test -- test/auth-paths.test.ts src/__tests__/graph-tools.test.ts`

## Caveats
- `npm run format:check` still reports pre-existing formatting drift across untouched files.
- Full `npm test` timed out in two tests during the whole-suite run on Windows; both timed-out files passed when rerun directly.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,31 @@ npx @softeria/ms-365-mcp-server --preset mail --list-permissions
 
 This is useful for enterprise environments where Graph API permissions must be pre-approved and admin-consented before deploying a new version.
 
+The `--list-permissions` JSON includes:
+
+- `permissions`: legacy alias for `toolPermissions`, kept for compatibility with existing scripts
+- `toolPermissions`: permissions implied by the currently enabled tool surface
+- `authScopes`: scopes the server will request during MSAL authentication
+- `missingAuthScopesForEnabledTools`: enabled tool permissions not covered by configured auth scopes
+- `extraAuthScopesNotImpliedByTools`: configured auth scopes that are not implied by the enabled tools
+
+### Explicit Auth Scopes
+
+By default, MSAL requests the same scopes implied by the enabled tools. Enterprise and headless deployments can override that authentication surface with `--auth-scopes` or `MS365_MCP_AUTH_SCOPES`:
+
+```bash
+npx @softeria/ms-365-mcp-server \
+  --org-mode \
+  --enabled-tools '^(list-mail-messages|get-mail-message|list-drives|get-drive-item|download-bytes)$' \
+  --auth-scopes 'User.Read Mail.Read Files.Read'
+```
+
+CLI value takes precedence over `MS365_MCP_AUTH_SCOPES`; if neither is set, the default tool-derived scope behavior is unchanged. Supplying an empty value fails at startup so deployments do not accidentally fall back to a wider scope set.
+
+`--auth-scopes` only changes the scopes requested from Microsoft. It does not hide tools. Use `--enabled-tools`, `--preset`, and `--read-only` to control which tools are exposed.
+
+In HTTP mode, OAuth discovery advertises the configured `authScopes` when `--auth-scopes` is set so clients request the same narrower consent surface. On-Behalf-Of mode (`--obo`) still advertises `api://<clientId>/access_as_user` for protected-resource metadata; `--auth-scopes` does not override OBO.
+
 ## Organization/Work Mode
 
 To access work/school features (Teams, SharePoint, etc.), enable organization mode using any of these flags:
@@ -474,6 +499,7 @@ The following options can be used when running ms-365-mcp-server directly from t
 --work-mode       Alias for --org-mode
 --force-work-scopes Backwards compatibility alias for --org-mode (deprecated)
 --cloud <type>    Microsoft cloud environment: global (default) or china (21Vianet)
+--auth-scopes <scopes> Override auth scopes requested from Microsoft (whitespace-separated)
 ```
 
 ### Server Options
--- a/README.md
+++ b/README.md
@@ -150,8 +150,6 @@ Scope coverage is hierarchy-aware: for example, `Mail.ReadWrite` covers tools th
 
 In HTTP mode, OAuth discovery advertises the effective filtered permissions so clients request the same consent surface. On-Behalf-Of mode (`--obo`) still advertises `api://<clientId>/access_as_user` for protected-resource metadata; `--allowed-scopes` does not override OBO.
 
-Durable headless token-cache storage and account pinning are separate concerns from tool-surface scope filtering and can be addressed independently.
-
 ## Organization/Work Mode
 
 To access work/school features (Teams, SharePoint, etc.), enable organization mode using any of these flags:
--- a/README.md
+++ b/README.md
@@ -123,28 +123,34 @@ This is useful for enterprise environments where Graph API permissions must be p
 
 The `--list-permissions` JSON includes:
 
-- `permissions`: legacy alias for `toolPermissions`, kept for compatibility with existing scripts
-- `toolPermissions`: permissions implied by the currently enabled tool surface
-- `authScopes`: scopes the server will request during MSAL authentication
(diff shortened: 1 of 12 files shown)


eirikb:
Hi there. Looks clean, but I just have to ask: Why? 😅

I get what the PR does, but I'm not sure what the point is. Or I get the point, but why would you need this?  
And additional scopes could probably be manually added to your own entra app.

castaples:
Thanks for asking. I should have explained the motivation more clearly.

The short version is: this isn’t about adding extra scopes to an Entra app. It’s almost the opposite. It’s for deployments where the Entra app is deliberately consented to a narrower scope set than the enabled tool surface would imply.

In our headless/enterprise deployment, Graph permissions are pre-approved by security/IT. We may expose or keep installed a broader tool surface, but only consent a narrower set of Graph scopes and let Graph return 403 for anything outside that policy boundary. Today the server derives the OAuth request scopes from the enabled tools/presets, so auth can fail before a tool call ever reaches Graph:

- HTTP/OAuth metadata and `/authorize` can ask the client/user/admin to consent to scopes we intentionally do not want.
- `acquireTokenSilent()` can request the full derived scope set and fail even though the cache has a valid token/refresh path for the narrower approved set.
- The workaround is to use BYOT/static access tokens or wrapper logic, which is more fragile than letting this server use normal MSAL refresh with an explicit approved scope set.

So the intended use case is:

```bash
ms-365-mcp-server \
  --org-mode \
  --enabled-tools '...' \
  --auth-scopes 'User.Read Mail.Read Files.Read'
```

`--enabled-tools` still controls what tools are exposed. `--auth-scopes` controls what MSAL asks Entra for. The new `--list-permissions` diagnostics are there to make the mismatch explicit, so operators can see which tools may 403 because the auth scope set is narrower.

You’re right that an operator *could* add the missing scopes to their own Entra app, but in our case that would be a broader grant than we want. We’re trying to avoid needing to ask for `Mail.Send`, `Mail.ReadWrite`, shared mailbox scopes, tenant-wide Teams scopes, etc. just because those tools exist in a preset or regex.

If this direction makes sense, there are two follow-up improvements I’d like to propose separately rather than bundle into this PR:

1. Strict account pinning for headless deployments, e.g. `MS365_MCP_EXPECTED_USERNAME` / `MS365_MCP_EXPECTED_HOME_ACCOUNT_ID`, so a cache with the wrong account fails fast instead of falling back to the first account.
2. A token-cache storage abstraction, so deployments can plug in durable cache stores without patching auth internals. We’d probably use that to back the MSAL cache with our platform’s secret store.

Happy to adjust this PR if you’d prefer a different name or shape for the option. The main goal is to separate “which tools are visible” from “which scopes this deployment is allowed to request”.

eirikb:
Aha, that changes things.

But your PR still allows the tools? So if you filter out Mail.Send, the tool send-mail is still available?  
It almost sounds like you could do with simply tool filtering.
Or - tool filtering based on scopes?

castaples:
Yes, that’s a fair challenge.

In the specific simple case, tool filtering is enough, and I don’t want to overstate the need for this. If a deployment can express its policy entirely as “only expose tools whose scopes are consented”, then `--enabled-tools` / presets are the cleaner answer.

Where I was trying to solve a slightly different problem is the enterprise/headless case where the operator wants the server to stay bootable and refreshable under a known approved auth scope set, while still making scope/tool mismatches visible in diagnostics instead of turning them into an auth-time failure.

That said, your “tool filtering based on scopes” idea might actually be the better abstraction.

Something like one of these would probably align more naturally with how the project already works:

```bash
# expose only tools whose required scopes are covered by this set
ms-365-mcp-server --allowed-scopes "User.Read Mail.Read Files.Read"

# or in addition to existing filters
ms-365-mcp-server \
  --preset mail,files \
  --allowed-scopes "User.Read Mail.Read Files.Read"
```

Then the server would:
- derive the normal tool permission set from endpoints;
- filter/hide tools whose required scopes are not covered by `allowedScopes`;
- request only the allowed/remaining scopes;
- keep `--list-permissions` useful by reporting hidden tools and missing scopes.

That would be safer than my current PR because the tool surface and auth surface would not intentionally diverge. It also maps better to the operator intent: “this app is allowed to use these scopes, expose whatever tools fit inside that boundary.”

Would you prefer I rework this PR in that direction? I’m happy to replace `--auth-scopes` with something like `--allowed-scopes` / `MS365_MCP_ALLOWED_SCOPES` that filters tools by scope coverage rather than allowing tools to remain visible and 403 later.

eirikb:
Yeah please do. Having a `--allowed-scopes` would make more sense.

The only problem is probably that one might be missing tools without understanding why. Not a big deal though, and logging should show when tools are flipped off because of the flag.

castaples:
Reworked this PR around the maintainer-requested `--allowed-scopes` approach:

- `--allowed-scopes` / `MS365_MCP_ALLOWED_SCOPES` now acts as a Graph tool-surface boundary instead of only changing requested OAuth scopes.
- Tool registration and discovery hide Graph tools whose endpoint scopes are not covered by the allowlist, with startup logs listing disabled tools and missing scopes.
- OAuth metadata and `/authorize` use the effective permissions for the filtered tool set; OBO protected-resource metadata still advertises `api://<clientId>/access_as_user`.
- `--list-permissions` now reports pre-filter tool permissions, effective permissions, allowed scopes, disabled tools, missing scopes, and unused allowed scopes.

I left durable headless token-cache storage and account pinning out of this PR as follow-up concerns.

eirikb:
Looks much better, thanks for reworking it 🙂

A few small things before merging:

- The README has a sentence about "Durable headless token-cache storage and account pinning" being separate concerns. Feels like leftover from this thread, probably doesn't belong in the user-facing docs.
- `isEndpointCoveredByAllowedScopes` is exported but I can't find any callers, looks like dead code.
- In `registerDiscoveryTools` you call `buildToolsRegistry` twice (unfiltered first for the log line, then filtered for actual use). Should be doable in one pass I think?

Otherwise looks good, the `--list-permissions` diagnostics are nice.

castaples:
Thanks, addressed these in `dc38772`:

- Removed the durable token-cache/account-pinning follow-up sentence from the README.
- Removed the unused `isEndpointCoveredByAllowedScopes` helper/export.
- Updated discovery registration so disabled allowed-scope diagnostics are collected during the same `buildToolsRegistry` pass used for registration.

Validation run:
- `npx prettier --write README.md src/auth.ts src/graph-tools.ts`
- `npm test -- test/allowed-scopes.test.ts src/__tests__/graph-tools.test.ts`
- `npm run lint` (0 errors; existing warnings only)
- `npm run build`

eirikb:

PR #479, opened 2026-05-17
fix: preserve @odata.deltaLink and strip $top from delta endpoint schemas

mikemrozek:
Two related bug fixes affecting all delta endpoints (mail, calendar, drive, sharepoint, etc.):

**1. `@odata.deltaLink` stripped from responses**

The response serializer was stripping `@odata.deltaLink` alongside other OData properties, making delta sync impossible — callers got the first page of items but no resume token. Fixed by whitelisting `@odata.deltaLink` in `removeODataProps`, mirroring the existing `@odata.nextLink` exception.

**2. `fetchAllPages` never captured `@odata.deltaLink`**

Even with the serializer fix, `fetchAllPages` never read the deltaLink because Graph only emits it on the final page, and the pagination loop only looked at `@odata.nextLink`. Fixed by reading deltaLink off each page during pagination and stamping the most recent one onto the combined response.

**3. `$top` exposed on delta tool schemas**

Delta tool schemas exposed a `top` parameter, but Graph rejects `$top` on delta endpoints with HTTP 400 (it requires the `Prefer: odata.maxpagesize` header instead). Stripping `top`/`$top` from delta tool param schemas eliminates the footgun. Belt-and-suspenders: also strip from outgoing queryParams for stale clients with cached schemas.

**Verified in production**

Running in production on our fork (v0.88.2-tsq.6). `list-mail-folder-messages-delta` with `fetchAllPages: true` now returns items plus `@odata.deltaLink` containing a valid `\$deltatoken` — verified end-to-end against a real mailbox.

**Affected delta tools:**
- `list-mail-folder-messages-delta`
- `list-calendar-events-delta`
- `list-calendar-view-delta`
- `get-drive-delta`
- `get-sharepoint-sites-delta`

References:
- [Graph delta query overview](https://learn.microsoft.com/en-us/graph/delta-query-overview)
- [Prefer: odata.maxpagesize header](https://learn.microsoft.com/en-us/graph/aad-advanced-queries)


eirikb:
Hi @mikemrozek

`@odata.deltaLink` fixes look great, thanks!

On the `$top` strip though - MS Learn actually documents `$top` as supported on three of the five endpoints we're stripping:

- `list-mail-folder-messages-delta`: https://learn.microsoft.com/en-us/graph/api/message-delta
- `get-drive-delta`: https://learn.microsoft.com/en-us/graph/api/driveitem-delta
- `get-sharepoint-sites-delta`: https://learn.microsoft.com/en-us/graph/api/site-delta

Only calendar/calendarView are silent on `$top`. Could you narrow the strip to those two before merging? Or if you saw a 400 on one of the others, would love to see the error - might be I'm reading the docs wrong.

eirikb:
@mikemrozek ping

PR #481, opened 2026-05-19
feat(auth): add pluggable auth cache storage

castaples:
## Summary
- Add a provider-neutral auth cache storage abstraction while preserving the default keytar/file behavior when no command is configured.
- Add `MS365_MCP_AUTH_CACHE_COMMAND` command storage for local MSAL token-cache and selected-account metadata, with durable save semantics and fail-closed sanitized command errors.
- Skip configured command storage for stateless HTTP request-token/OBO startup and Graph requests unless local auth tools or account commands explicitly use local auth state.

## Validation
- `npm test -- test/token-cache-storage.test.ts test/auth-storage.test.ts test/cache-stamp.test.ts test/auth-paths.test.ts test/auth-pinning.test.ts test/multi-account.test.ts test/auth-tools.test.ts test/cli.test.ts`
- `npm run lint` (passes with existing warnings)
- `npm run build`
- `npx prettier --check src/token-cache-storage.ts src/auth.ts src/index.ts src/startup-pinning.ts test/token-cache-storage.test.ts test/auth-storage.test.ts test/cli.test.ts README.md docs/deployment.md`
- `npm test` currently times out locally in the pre-existing `src/__tests__/graph-tools.test.ts` `$count advanced query mode` test when the full suite runs; rerunning that single test passes locally.

## Notes
- The command env var is an executable wrapper path only. It is not shell parsed, and v1 does not add an args env var or CLI flag.
- Command storage stores the existing stamped envelope string as an opaque payload so `_cacheEnvelope`, saved-at ordering, legacy raw cache compatibility, and selected-account storage remain compatible.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,3 +52,14 @@ MS365_MCP_TENANT_ID=common
 #   - Environment variables (AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_TENANT_ID)
 #
 # See README.md for detailed Azure Key Vault setup instructions.
+
+# -------------------------------------------------------------------
+# External Auth Cache Storage (Optional)
+# -------------------------------------------------------------------
+# Headless local-MSAL deployments can store token-cache and selected-account
+# metadata with a provider-neutral executable wrapper. The value is a real
+# executable path, not a shell command string; put any provider-specific args
+# inside the wrapper script.
+#
+# MS365_MCP_AUTH_CACHE_COMMAND=/path/to/ms365-auth-cache-store
+# MS365_MCP_AUTH_CACHE_COMMAND_TIMEOUT_MS=10000
--- a/README.md
+++ b/README.md
@@ -571,6 +571,8 @@ Environment variables:
 - `MS365_MCP_KEYVAULT_URL`: Azure Key Vault URL for secrets management (see Azure Key Vault section)
 - `MS365_MCP_TOKEN_CACHE_PATH`: Custom file path for MSAL token cache (see Token Storage below)
 - `MS365_MCP_SELECTED_ACCOUNT_PATH`: Custom file path for selected account metadata (see Token Storage below)
+- `MS365_MCP_AUTH_CACHE_COMMAND`: External executable wrapper for provider-neutral auth-cache storage (see Token Storage below)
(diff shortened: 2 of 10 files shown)


eirikb:
Hi @castaples, thanks for this! The refactor part is a nice cleanup, `auth.ts` is a lot easier to follow now and the atomic file write is a real improvement.

Two small things before merging:

1. In `runCommand` in `token-cache-storage.ts`, there's no error handler on `child.stdin`. If the wrapper exits before consuming stdin, `end(payload)` will emit EPIPE and (AFAIK) crash the parent in Node 18+. At least we've had similar issues with that before (yes we still support it and if  this is all we need to keep on doing so why not). A `child.stdin.on('error', () => {})` should be enough.
2. In `AuthManager.create()` the fallback creates storage without `allowCommandStorage: false`. Nothing hits that path today, but if a future caller forgets to pass storage explicitly they'd silently re-enable command storage in HTTP mode. Just defensive.

Also curious: was there a specific deployment that motivated this? I'm a bit on the fence about adding a published command protocol when `MS365_MCP_TOKEN_CACHE_PATH` already covers most headless cases (mount a volume, sync via sidecar, etc). Not a no, just trying to understand the use case before we commit to the protocol shape.

Thanks 🙂

castaples:
Thanks, agreed on both points. I pushed `c453465` with a `child.stdin` error handler in the command runner and made the `AuthManager.create()` fallback explicitly disable command storage. I also added focused regression tests for the stdin error case and the fallback path.

On the deployment motivation: our concrete case is a headless AWS ECS task. The local container filesystem is ephemeral, and we would rather not mount a shared volume solely for the MSAL cache. The durable store we already approve for this deployment shape is AWS Secrets Manager, so the command protocol lets us bridge to that store without making upstream take an AWS dependency.

I agree `MS365_MCP_TOKEN_CACHE_PATH` remains the right simpler option when a deployment already has a persistent volume, sidecar sync, or another local durable filesystem pattern. The command provider is mainly for the no-local-durable-filesystem case, and the protocol stays provider-neutral so the same hook could bridge to a database, object store, secrets manager, or similar external store.

PR #482, opened 2026-05-19
feat(http): add --trust-proxy-auth to skip Bearer check on /mcp

salmon-21:
## Summary
Adds an opt-in `--trust-proxy-auth` CLI flag that disables the in-process Bearer-token middleware on `GET/POST /mcp` so the server can run unauthenticated behind a reverse proxy that has already authenticated the caller.

When the flag is set, Microsoft Graph API access still flows through the locally cached MSAL refresh token via `AuthManager` — the same path stdio mode uses today — so a one-time device-code login still bootstraps the token cache. The OAuth metadata endpoints (`/register`, `/token`, `/authorize`, `/.well-known/*`) are unchanged; they simply go unused when nothing reaches them through the proxy.

The flag is **off by default**. The README guidance ("HTTP mode requires authentication") still holds for any deployment that is reachable from anything other than a trusted proxy.

## Motivation
Self-hosted multi-backend MCP setups (claude.ai → gateway → N MCP backends) typically authenticate at the gateway. Today the only way to put ms-365-mcp-server behind such a gateway is to wrap stdio in supergateway, which has been observed to crash on mid-request client disconnects under recent SDK versions. A documented opt-out lets the gateway own auth and ms-365-mcp-server stay native HTTP.

## Changes
- src/cli.ts: new --trust-proxy-auth option; CommandOptions.trustProxyAuth.
- src/server.ts: when the flag is set, substitute a pass-through middleware for microsoftBearerTokenAuthMiddleware on the two /mcp routes. The existing if (req.microsoftAuth) { ... } else { await handler(); } branch already handles the no-bearer case, so the handler body itself does not change.

## Test plan
- [x] npm install && npm run generate && npm run build
- [x] node dist/index.js --http :3001 --trust-proxy-auth listens on 3001 with no MSAL login required at startup
- [x] POST /mcp tools/list returns 200 + tool list without an Authorization header
- [x] node dist/index.js --http :3001 (no flag) still returns 401 + WWW-Authenticate on the same request
- [x] Docker image built from the branch successfully fronts ToDo tool calls via an external reverse proxy

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -82,6 +82,10 @@ program
     '--obo',
     'Enable On-Behalf-Of token exchange in HTTP mode. Exchanges the incoming bearer token for a Graph API token using the OBO flow. Requires MS365_MCP_CLIENT_SECRET.'
   )
+  .option(
+    '--trust-proxy-auth',
+    'In HTTP mode, skip the built-in Bearer-token check on /mcp and let an upstream reverse proxy (which is presumed to have already authenticated the caller) pass requests through unauthenticated. Microsoft Graph access still flows through the locally cached MSAL refresh token via AuthManager. Use only when the server is not reachable except via a trusted proxy.'
+  )
   .addOption(
     // DEPRECATED: kept only so existing deployments that set --base-url or
     // MS365_MCP_BASE_URL do not crash at startup. Use --public-url /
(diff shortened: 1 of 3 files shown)


eirikb:
Hi @salmon-21

Thanks for the PR! Verified the AuthManager fallback works as you described.

Small notes, not blockers:

- The CLI help doesn't mention that any forwarded `Authorization` header gets dropped when the flag is on, so all callers share the locally cached identity. Worth a line.
- `--obo` + `--trust-proxy-auth` together silently turn OBO into a no-op. Could error at startup the way `--obo` already does for `!--http`.
- An `MS365_MCP_TRUST_PROXY_AUTH` env var would fit the pattern of the other deployment flags, given the Docker motivation.

I'll merge as is and follow up on these myself. Thanks :slightly_smiling_face:

PR #499, opened 2026-05-29
feat: add Planner task chat message tools

lestephen:
## Why this matters now

Microsoft is retiring the classic Planner task "comments" experience and replacing it with task chat as part of the early-2026 Planner update (Message Center post MC1193421). The old comments UI is being hidden from the task details pane, while task chat becomes the supported in-task collaboration surface.

For an MCP client that needs to post per-task updates teammates will actually see, Planner task chat is the current surface. Today this server can only reach the legacy `plannerTask.conversationThreadId` comments indirectly through generic group-thread tools, and it has no task-chat tools.

## What this adds

Adds three Planner task chat tools backed by the Graph beta Planner messages API:

| Tool | Method + path | Scope |
| --- | --- | --- |
| `list-planner-task-messages` | `GET /planner/tasks/{id}/messages` | `Tasks.Read` |
| `create-planner-task-message` | `POST /planner/tasks/{id}/messages` | `Tasks.ReadWrite` |
| `delete-planner-task-message` | `DELETE /planner/tasks/{id}/messages/{messageId}` | `Tasks.ReadWrite` |

The implementation now builds on the generator beta support merged upstream, rather than carrying the older per-endpoint beta-routing implementation from the original PR.

## Generator note

While rebasing onto the new beta generator path, the generated beta client was still emitting `z.void()` for these successful Planner message responses because the beta OpenAPI metadata uses wildcard `2XX` response codes. This PR normalizes wildcard success responses in the simplified OpenAPI output so `openapi-zod-client` generates usable response schemas for the selected beta endpoints.

## Beta caveats

The tool tips call out the Graph beta constraints:

- beta APIs are subject to change
- delegated work/school accounts only
- global cloud only

## Tests

- `test/endpoints-validation.test.ts` confirms the Planner task chat tools are generated and that list/create have non-void beta response schemas.
- Full `npm run verify` passes: generate, lint, format check, build, and tests.

### DIFF
--- a/bin/modules/simplified-openapi.mjs
+++ b/bin/modules/simplified-openapi.mjs
@@ -145,6 +145,10 @@ export function createAndSaveSimplifiedOpenAPI(
     }
   }
 
+  if (apiVersion === 'beta') {
+    normalizeWildcardSuccessResponses(openApiSpec.paths);
+  }
+
   if (openApiSpec.components && openApiSpec.components.schemas) {
     removeODataTypeRecursively(openApiSpec.components.schemas);
     flattenComplexSchemasRecursively(openApiSpec.components.schemas);
@@ -160,6 +164,29 @@ export function createAndSaveSimplifiedOpenAPI(
   pruneUnusedSchemas(openApiSpec, usedSchemas);
 
   fs.writeFileSync(openapiTrimmedFile, yaml.dump(openApiSpec));
+}
+
+function normalizeWildcardSuccessResponses(paths) {
+  Object.values(paths || {}).forEach((pathItem) => {
+    if (!pathItem || typeof pathItem !== 'object') return;
+
+    Object.entries(pathItem).forEach(([method, operation]) => {
+      if (!operation || typeof operation !== 'object') return;
+      if (!operation.responses || !operation.responses['2XX']) return;
+
+      const hasConcreteSuccess = Object.keys(operation.responses).some((statusCode) =>
+        /^2\d\d$/.test(statusCode)
+      );
+      if (hasConcreteSuccess) {
+        delete operation.responses['2XX'];
+        return;
+      }
+
+      const successStatus = method === 'post' ? '201' : method === 'delete' ? '204' : '200';
+      operation.responses[successStatus] = operation.responses['2XX'];
+      delete operation.responses['2XX'];
+    });
+  });
 }
 
 function removeODataTypeRecursively(obj) {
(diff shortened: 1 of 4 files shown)


lestephen:
Added a **"Why this matters now"** section to the description to make the motivation explicit:

Microsoft is retiring the classic Planner task **comments** experience and replacing it with **task chat** in the early-2026 Planner update ([MC1193421](https://mc.merill.net/message/MC1193421), rolling out mid-Jan–mid-Feb 2026). The old comments UI is being **hidden from the task** (the details pane just shows a link to view them in Outlook), while task chat becomes the supported way for teammates to collaborate on a task with rich text and @mentions.

This server can currently only reach the *deprecated* comments indirectly (via the generic group-thread tools) and has **no way to use task chat at all** — so as the old UI disappears, this is the gap between per-task updates teammates actually see vs. updates buried behind an Outlook link. That's the case for adding these tools now, despite the API still being `/beta`-only.

lestephen:
Hi! Quick maintainer-direction check on this and the related PRs (#500, #501).

We realize these may not be on your roadmap, and no pressure if they are better kept in our fork. Before we spend more time rebasing/refining them, would you be open to this general direction upstream?

If the concern is scope, maintenance burden, API shape, or splitting, we are happy to adjust. A simple "yes, but needs changes," "not right now," or "fork-only is best" would be totally useful.

eirikb:
Hi

Thanks for checking in, and sorry for the slow reply. 
I'm open for this direction, but I think the correct approach is beta support in the generator.
I think I've mentioned this before - and it might sound scary, but perhaps it's not that complicated.  
And doing so will probably help your other PRs...

lestephen:
Thanks, that makes sense. We'll take a look at making beta support part of the generator first, then rework these PRs on top of that instead of carrying the beta endpoints manually.

To confirm the shape: would you prefer the generator to support per-endpoint `apiVersion` metadata in `endpoints.json`, downloading/merging the beta OpenAPI metadata only for endpoints marked `beta`? If that sounds right, we can start with a small generator PR before revising #499/#500/#501.

eirikb:
Hi @lestephen

Perhaps we can pick up this now with the new beta stuff? I just merged it and from a quick test it looks like it will work.

lestephen:
Thanks! I rebased this onto current `main` and reworked it to use the new generator beta support rather than the older per-endpoint beta-routing implementation.

The PR is now narrower:

- adds the Planner task chat endpoints in `endpoints.json`
- regenerates the beta client through the upstream generator path
- includes a small generator fix so beta `2XX` responses produce usable response schemas for these tools
- keeps validation coverage for the generated beta tool surface

I also cleaned this down to a single commit and updated the PR description so it reflects the current generator-backed approach. `npm run verify` passes locally.

eirikb:

PR #500, opened 2026-05-29
feat: add get-download-url tool for drive downloads

lestephen:
## Problem

`download-bytes` returns binary content as base64 inline in the tool response. For anything beyond small files, that is expensive and fragile: the bytes round-trip through the model or agent context and can be truncated by client output limits.

Drive and SharePoint files already expose Microsoft Graph's native short-lived `@microsoft.graph.downloadUrl`, which lets the client stream bytes directly to disk without an `Authorization` header and without base64 in the tool response.

## What this adds

Adds a read-only utility tool, `get-download-url`, that resolves `@microsoft.graph.downloadUrl` for OneDrive and SharePoint drive items.

Supported target shapes are explicit drive item metadata or content paths, for example:

- `/drives/{drive-id}/items/{driveItem-id}/content`
- `/me/drive/items/{driveItem-id}`
- `/sites/{site-id}/drive/items/{driveItem-id}`
- `/me/drive/root:/path/file.ext:/content`

The tool strips a true drive item `/content` suffix before reading item metadata, because `@microsoft.graph.downloadUrl` lives on the drive item metadata rather than the `/content` response.

## Guardrails

- Rejects query-shaped targets rather than silently changing request semantics.
- Rejects unsupported non-drive Graph paths before making an authenticated Graph request.
- Rejects mail/calendar attachment paths and other `/$value` byte endpoints with guidance to use `download-bytes`.
- Treats meeting recordings as authenticated byte reads, not pre-authenticated URL resources.
- Surfaces underlying Graph errors instead of masking them as "no download URL available."

## Notes

- Purely additive; no behavior change to existing tools.
- Read-only (`readOnlyHint: true`).
- Updates the `download-bytes` and initialization guidance so large drive/SharePoint file reads prefer `get-download-url`, while mail attachments, profile photos, Teams hosted content, and meeting recordings stay on authenticated byte reads.

## Tests

Adds coverage for successful drive/SharePoint URL resolution, `/content` normalization, query rejection, unsupported Graph path rejection, mail/calendar/group attachment rejection, meeting recording rejection, Graph error passthrough, bearer account mismatch handling, and the updated tool guidance.

Full `npm run verify` passes: generate, lint, format check, build, and tests.

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -128,6 +128,11 @@ function createMockServer() {
   };
 }
 
+function makeJwt(payload: Record<string, unknown>): string {
+  const body = Buffer.from(JSON.stringify(payload)).toString('base64url');
+  return `header.${body}.signature`;
+}
+
 // ========== TESTS ==========
 
 describe('graph-tools', () => {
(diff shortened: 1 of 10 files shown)


eirikb:
Hi @lestephen

Same as #499 here - a rebase on main with the new beta stuff should help this one too 🙂

lestephen:
Thanks, agreed. I rebased this onto current `main` as well.

This PR is now independent of the old beta-routing work and is limited to a read-only `get-download-url` utility for Graph resources that expose `@microsoft.graph.downloadUrl`. During the rework I tightened the contract so it only accepts explicit OneDrive/SharePoint drive item paths, rejects query-shaped targets, and leaves mail/calendar attachments plus meeting recordings on authenticated byte reads via `download-bytes`.

I also cleaned this down to a single commit and updated the PR description, including removing the earlier recording-download-url wording. `npm run verify` passes locally.

eirikb:
Could you fix the conflict? After that we'll merge

lestephen:
Thanks, and sorry about that. I forgot to check the PR sequencing after #499 merged, so #500 still had a small conflict in `test/endpoints-validation.test.ts`.

I rebased #500 onto current `main` and resolved the conflict by keeping both the Planner beta response-schema assertions from #499 and the `get-download-url` endpoint-guidance assertions from this PR.

While doing that, I ran a second review pass and caught one small issue we missed in the original review: discovery-mode search still preferred `download-bytes` for drive/SharePoint file-download queries, even though this PR is trying to route those larger file reads through `get-download-url`. I fixed that by adding explicit discovery keywords for the utility tool and golden-query coverage for `download drive file`, `sharepoint file download`, and `large drive file`.

`npm run verify` passes locally.

eirikb:

PR #514, opened 2026-06-06
feat(http): allow MCP discovery requests without a bearer token

### LINKED ISSUE #513: Allow MCP discovery requests (tools/list, etc.) without a bearer token, for gateway pre-auth catalog sync
### Problem / use case

We run this server in HTTP mode behind our own MCP **gateway** (one backend per app, e.g. `--preset outlook`). The gateway needs to do two things that the current auth options can't both satisfy at once:

1. **Enumerate the tool catalog right after deployment, before any user has authenticated.** The gateway calls `tools/list` to discover which tools the backend exposes, syncs that catalog, and applies its own secondary policy on top (which tools to surface to end users).
2. **Inject a per-user identity on actual calls.** Once the gateway has authenticated the user, it injects that user's managed OAuth (Microsoft) access token into the request, which flows through to this server so Graph calls run as that user.

Neither current mode supports both:

- **Default HTTP**: `tools/list` returns `401` without a token, so the gateway can't enumerate tools pre-auth.
- **`--trust-proxy-auth`**: lets tokenless requests through, but it skips the bearer entirely and falls back to a single locally-cached MSAL identity — so the per-user token the gateway injects is ignored, and every user shares one account.

### Proposed solution

Allow the MCP **discovery** JSON-RPC methods through the bearer middleware without a token, while everything else (notably `tools/call`) still requires a valid bearer that is passed through to Graph:

- tokenless allow-list: `initialize`, `notifications/initialized`, `tools/list`, `prompts/list`, `resources/list`, `ping`
- `tools/call` with no token → still `401`
- `tools/call` with a bearer → passed through per-request (existing behavior)

This is complementary to `--trust-proxy-auth`: that flag is for "gateway owns auth, single shared identity"; this is for "gateway enumerates tools pre-auth, then injects per-user tokens for calls."

### Notes

I already have a small implementation + tests and would be happy to open a PR (adds an `isDiscoveryRequest` check in `microsoftBearerTokenAuthMiddleware`, ~100 lines incl. tests). Happy to gate it behind a flag (e.g. `--allow-unauthenticated-discovery`) if you'd prefer it opt-in rather than default.

feifeigood:
## Summary

Lets the HTTP bearer-token middleware pass **MCP discovery requests** through without a token, so an MCP gateway can enumerate a backend's tool catalog before any user has authenticated. `tools/call` still requires a valid bearer, which is passed through per-request to Graph as today.

Tokenless allow-list (`isDiscoveryRequest`): `initialize`, `notifications/initialized`, `tools/list`, `prompts/list`, `resources/list`, `ping`.

## Motivation

Running behind an MCP gateway (one backend per app via `--preset`), we need to:
1. **List tools right after deploy, pre-auth** — the gateway syncs the catalog and applies its own secondary policy on which tools to expose.
2. **Inject a per-user identity on calls** — after the gateway authenticates the user, it injects that user's managed OAuth Microsoft token into the request.

Neither existing mode does both: default HTTP `401`s `tools/list` without a token; `--trust-proxy-auth` allows tokenless requests but ignores the per-request bearer and uses a single cached MSAL identity. This change is complementary to `--trust-proxy-auth`. See #513.

## Changes

- `src/lib/microsoft-auth.ts`: add `isDiscoveryRequest()` and let discovery methods skip the missing-bearer `401` in `microsoftBearerTokenAuthMiddleware`. Non-discovery requests are unchanged; a present bearer is still parsed/validated and passed through.
- `test/discovery-auth.test.ts`: covers discovery-allowed, `tools/call` `401` without token, per-user passthrough with a bearer, and `trustProxyAuth` short-circuit.

## Notes

This is **on by default** (discovery only ever exposes tool schemas, never data; calls still require auth). If you'd rather it be opt-in, I'm happy to gate it behind a flag like `--allow-unauthenticated-discovery` — just say the word.

Closes #513

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -86,6 +86,10 @@ program
     '--trust-proxy-auth',
     'In HTTP mode, skip the built-in Bearer-token check on /mcp and ignore any forwarded Authorization header. All callers share the locally cached MSAL identity (same path stdio mode uses). Use only when an upstream reverse proxy has already authenticated the caller.'
   )
+  .option(
+    '--allow-unauthenticated-discovery',
+    'In HTTP mode, allow MCP discovery requests (initialize, tools/list, prompts/list, resources/list, ping) without a bearer token, so a gateway can enumerate the tool catalog before any user has authenticated. Non-discovery requests (e.g. tools/call) still require a token. Off by default.'
+  )
   .addOption(
     // DEPRECATED: kept only so existing deployments that set --base-url or
     // MS365_MCP_BASE_URL do not crash at startup. Use --public-url /
@@ -122,6 +126,7 @@ export interface CommandOptions {
   authBrowser?: boolean;
   obo?: boolean;
   trustProxyAuth?: boolean;
+  allowUnauthenticatedDiscovery?: boolean;
   publicUrl?: string;
   /** @deprecated use publicUrl */
   baseUrl?: string;
@@ -266,6 +271,13 @@ export function parseArgs(): CommandOptions {
     process.env.MS365_MCP_TRUST_PROXY_AUTH === '1'
   ) {
     options.trustProxyAuth = true;
+  }
+
+  if (
+    process.env.MS365_MCP_ALLOW_UNAUTHENTICATED_DISCOVERY === 'true' ||
+    process.env.MS365_MCP_ALLOW_UNAUTHENTICATED_DISCOVERY === '1'
+  ) {
+    options.allowUnauthenticatedDiscovery = true;
   }
 
   // Handle cloud type - CLI option takes precedence over environment variable
--- a/src/lib/microsoft-auth.ts
+++ b/src/lib/microsoft-auth.ts
@@ -24,6 +24,24 @@ function isJwtExpired(token: string): boolean {
   }
 }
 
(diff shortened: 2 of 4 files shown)


eirikb:
Hi, and thanks for this 🙂

As you say, perhaps this shouldn't be on by default.
Could you gate it behind `--allow-unauthenticated-discovery` (default off)? Then it's opt-in and your gateway just sets the flag.

Other than that it LGTM. Thanks again 🙂

feifeigood:
Thanks for the review! 🙏
Done — it's now gated behind `--allow-unauthenticated-discovery` (default off), so it's fully opt-in. With the flag unset, discovery requests get a 401 like any other unauthenticated request; with it set, `initialize`/`tools/list`/`prompts/list`/`resources/list`/`ping` are allowed without a token while `tools/call` still requires a valid bearer. Also added an `MS365_MCP_ALLOW_UNAUTHENTICATED_DISCOVERY` env var for parity with the other flags, and updated the regression tests to cover both the flag-on and flag-off (default) paths.

The branch has been force-pushed with the updated single commit. PTAL 🙂

eirikb:
LGTM. Thanks :slightly_smiling_face:

PR #519, opened 2026-06-09
feat(config): bound fetchAllPages with MS365_MCP_MAX_PAGES and MS365_MCP_MAX_ITEMS

okapi-ca:
## What
Make the `fetchAllPages` pagination caps configurable and add an off switch.

## Why
When a tool is called with `fetchAllPages: true`, `graph-tools.ts` follows `@odata.nextLink` with hardcoded limits (`maxPages = 100`, `maxItems = 10_000`). Operators running this server with concurrent users may want a smaller per-request memory footprint (10K items × ~5KB ≈ 50MB), or to disable multi-page following entirely for security-conscious workloads. This mirrors the existing `MS365_MCP_MAX_TOP` pattern.

## How
Three new env vars (invalid values warn and fall back to the default):
- `MS365_MCP_MAX_PAGES` — default `100`
- `MS365_MCP_MAX_ITEMS` — default `10000`
- `MS365_MCP_ALLOW_PAGINATION=0|false|no` — disables nextLink following; `fetchAllPages: true` then returns the first page only

## Backward compatibility
Defaults match the previous hardcoded values — no behaviour change unless explicitly configured. Adds 3 unit tests and documents the vars in the README.


eirikb:
Thanks :)

PR #520, opened 2026-06-09
feat(security): helmet headers and per-IP rate limiting for HTTP mode

okapi-ca:
## What
Add `helmet` security headers and per-IP `express-rate-limit` to the HTTP-mode Express app.

## Why
HTTP mode exposes the OAuth surface (`/authorize`, `/token`, `/register`) and the `/mcp` endpoint with no security headers and no flood protection.

## How
- **helmet** — security headers on every response. CSP disabled (this server returns JSON / OAuth metadata, not HTML), HSTS enabled (TLS terminated upstream).
- **express-rate-limit** (opt out with `MS365_MCP_RATE_LIMIT_DISABLED=true`) — 30 req/min per IP on the auth surface (one shared bucket across the three routes), 120 req/min on `/mcp`. IETF draft-7 RateLimit headers.

## Behavioural note — `trust proxy`
The previous `trust proxy: true` trusts every hop, which lets a client spoof the leftmost `X-Forwarded-For` entry and bypass per-IP limiting (`express-rate-limit` `ERR_ERL_PERMISSIVE_TRUST_PROXY`). The default is now a single hop. `MS365_MCP_TRUST_PROXY_HOPS=<n>` covers multi-hop chains, `0` disables proxy trust, and a comma-separated subnet list is accepted. **Single-proxy deployments (the common case) are unaffected; multi-hop deployments must now set this var.**

## Backward compatibility
helmet + rate-limit are additive (rate-limit opt-out via env). The `trust proxy` default change is the one behavioural shift — called out above for review. Adds dependencies `helmet ^8.1.0` and `express-rate-limit ^7.5.1`, a rate-limit smoke test (no new test dependency — native `fetch`), and README docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,6 +574,8 @@ Environment variables:
 - `MS365_MCP_OUTPUT_FORMAT=toon`: Enable TOON output format (alternative to --toon flag)
 - `MS365_MCP_MAX_TOP=<n>`: Hard cap for Graph `$top` / `top` on list requests (positive integer). When the model passes a larger value, the server clamps it to `n` so responses stay smaller. Example: `MS365_MCP_MAX_TOP=15`
 - `MS365_MCP_BODY_FORMAT=html`: Return email bodies as HTML instead of plain text (default: text)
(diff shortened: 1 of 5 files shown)


eirikb:
Thanks :)

The more of these things we add I keep re-thinking "don't put this public, put it behind a good reverse proxy" would have been better. But oh well, here we are.

Issue #522, opened 2026-06-09
Silent token refresh fails ~1h after device-code login (stdio + personal Microsoft account, 0.114.5)


koopatroopa:
## Summary

On **stdio transport with a personal Microsoft account** (device-code flow), silent token refresh fails roughly **one hour after login**. The server works fine until the access token expires, then every tool call fails with:

```
Login failed: Silent token acquisition failed
Failed to acquire token for account '<msa>'. The token may have expired. Please re-login with: --login
```

Re-logging in restores service for ~another hour, then the same death. This makes unattended use (cron) effectively impossible — the token is almost never alive when a scheduled job fires.

## Environment

- **Version:** 0.114.5 (latest; no newer/pre-release exists)
- **Transport:** stdio (`npx -y @softeria/ms-365-mcp-server`, no flags, empty env, default shared public client)
- **Account:** personal Microsoft account (outlook.com / MSA), device-code auth
- **OS:** Linux, token cache in gnome-keyring via keytar

## Confirmed reproduction (on 0.114.5)

- Clean device-code login at 14:42 BST → `verify-login` success, Graph calls work.
- Same login at 15:54 BST (72 min later, no further interaction) → `verify-login` returns `Silent token acquisition failed`; `get-current-user` returns `Failed to acquire token... token may have expired`.
- A separate clean login the previous evening behaved identically: worked for ~8 minutes, dead by next morning.

## Timeline

- Token had been stable and auto-renewing for ~5 weeks (cache created 27 Apr).
- First failure: 5 Jun 20:40, while running 0.114.0 (which had been stable 8 days) — so a version bump didn't trigger it.
- Daily silent-acquisition failures since: 5, 6, 7, 8, 9 Jun.
- One `invalid_grant` logged: 7 Jun 00:37 (`Error in device code flow: post_request_failed: invalid_grant`).

## What is NOT the cause (ruled out)

- **Not a stale version** — already on latest 0.114.5; no newer/pre-release tag.
- **Not missing persistence** — the keytar cache IS written on login (encrypted `_cacheEnvelope` / `data` ~7.2 KB / `savedAt`), so a refresh token is being stored. This differs from the older closed reports #257 (cache not persisted) and #298 (keytar/file desync) — here persistence works, yet silent refresh still fails.
- **Not documented account revocation** — no password change / admin revoke (the AADSTS50173 triggers) occurred.

## Suspected cause

[...]

Issue #486, opened 2026-05-21
Server crashes silently due to missing global unhandledRejection handler — token refresh kills process after idle


oleksandrilyin:
## Bug Description

The server crashes silently during idle periods (~40-60 minutes after last use) with no preceding error in the Claude Desktop log:

```
Server transport closed unexpectedly, this is likely due to the process exiting early
Server disconnected.
```

## Root Cause

`src/index.ts` has no global `process.on('uncaughtException')` or `process.on('unhandledRejection')` handlers.

When MSAL triggers a background token refresh and the refresh throws an unhandled promise rejection (network hiccup, Keychain access delay, etc.), Node.js kills the entire process immediately with no log entry.

The crash is **not triggered by any user API call** — it happens during idle, which makes it hard to reproduce on demand.

## Evidence from logs

- Last successful API call at `08:50 UTC`
- - Server crash at `09:29 UTC` — 39 minutes later, zero client requests in between
- - - No error message logged before the crash (process killed before any handler runs)
- - - - Pattern repeated multiple times: server dies ~40 min after last activity
## Environment

- `@softeria/ms-365-mcp-server` v0.111.0
- - Claude Desktop, macOS 26.5 (Apple M5)
- - - `--org-mode`, stdio transport
## Suggested Fix

Add global handlers at the top of `src/index.ts` before `main()`:

```typescript
process.on('uncaughtException', (err: Error) => {
  console.error('[ms365-mcp] Uncaught exception (server staying alive):', err.message, err.stack);
});

process.on('unhandledRejection', (reason: unknown) => {
  console.error('[ms365-mcp] Unhandled rejection (server staying alive):', reason);
});
```

This keeps the server alive and logs the actual error instead of silently dying. The token refresh failure becomes visible and non-fatal.

## Impact

Every user running in stdio mode with `--org-mode` will experience random disconnects every 40-60 minutes of idle time. Claude Desktop shows "Server disconnected" and requires manual restart.

eirikb:
@oleksandrilyin See #493

eirikb:
Hopefully fixed by #493

PR #523, opened 2026-06-09
fix(auth): log MSAL error code when silent token acquisition fails

koopatroopa:
Both `acquireTokenSilent` call sites in `AuthManager` (`getToken` and `getTokenForAccount`) caught the MSAL failure with a bare `catch {}` and rethrew a generic message, discarding the `AuthError`. That makes the most common production failure (#522, and the silent-death symptom behind #486) undiagnosable from logs: there is no way to tell an upstream `invalid_grant` from `interaction_required` or a local cache miss.

Bind the error and log its `errorCode`, `subError` and `correlationId` via a small `describeAuthError` helper. The process-level handlers added in #493 do not cover this path, because the error is caught rather than thrown unhandled.

Example output on the stdio device-code path with a personal Microsoft account:

```
Silent token acquisition failed: invalid_grant / bad_token (correlationId: ...): AADSTS70000: The user could not be authenticated as the grant is expired.
```

Diagnostics only; no change to token behaviour. Refs #522.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,5 +1,5 @@
 import type { AccountInfo, Configuration } from '@azure/msal-node';
-import { PublicClientApplication } from '@azure/msal-node';
+import { AuthError, PublicClientApplication } from '@azure/msal-node';
 import logger from './logger.js';
 import { readFileSync } from 'fs';
 import { fileURLToPath } from 'url';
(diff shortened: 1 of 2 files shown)


eirikb:
Ooo nice, thanks. Showing the error instead of swallowing is good.

Would be great if we could fix 522 as well - perhaps trigger re-login (except that won't work well on stdio...)?

PR #528, opened 2026-06-10
fix(auth): suggest the consumers authority when an MSA refresh token is rejected on common

koopatroopa:
## What

When silent token acquisition fails with `invalid_grant` for a personal Microsoft account on the default `common` authority, the error and log now tell the user to set `MS365_MCP_TENANT_ID=consumers` and re-login, instead of the generic "token may have expired". README documents the same on the env var.

## Why

Follow-up to #522. As of June 2026, MSA refresh tokens issued via `/common` are rejected at the first refresh (`invalid_grant` / AADSTS70000), so stdio sessions die roughly an hour after login - that's the access-token lifetime, with the refresh failing underneath. The same login via `/consumers` keeps refreshing fine: on the default bundled client, two cache-verified refreshes succeeded at 61 minutes and 2h14m post-login, where `common` failed at its first refresh.

The generic message sends affected users in circles: re-logging in via `common` just restarts the same one-hour clock.

(One subtlety for reviewers: MSAL's proactive-refresh path swallows 4xx refresh failures, so the failure - and therefore the hint - only surfaces at hard expiry. Consistent with the ~1h observed timing.)

## Scope

- Covers the two `acquireTokenSilent` catch sites in `src/auth.ts` - the only emitters of the generic advice. Everything user-visible in stdio mode (Graph calls, `--verify-login`, the login/verify-login tools) funnels through them.
- Known sibling not covered: the HTTP/OAuth `/token` refresh path (`src/lib/microsoft-auth.ts`) surfaces the same upstream failure as a raw RFC 6749 error body to machine clients; the README note applies to that mode too. Happy to follow up if wanted.
- Considered auto-switching MSA accounts to `/consumers` on the silent request, left out deliberately: I've only verified refreshing a token *issued* via `/consumers`; refreshing a `/common`-issued RT against `/consumers` is untested, and guessing wrong there would be worse than a hint. A proactive warning at login time (MSA + `common` detected) would prevent the hour of confusion rather than explain it afterwards - also happy to follow up with either shape if you'd prefer.

## Testing

- New tests in `test/auth-silent-error.test.ts`: the helper's guard conditions (authority anchoring, casing, MSA-vs-work tenant, error-code filtering) plus both call-site compositions, asserting the rendered error messages
- `npm run verify` green: generate, eslint, prettier --check, tsup, 422 tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,7 +584,7 @@ Environment variables:
 - `SILENT=true|1`: Disable console output
 - `MS365_MCP_REDACT_PII=true|1`: Scrub JWTs, Bearer headers, OAuth token fields, and email addresses from log messages before they are written (default: disabled). Useful when logs are shipped to a central store or shared host.
 - `MS365_MCP_CLIENT_ID`: Custom Azure app client ID (defaults to built-in app)
-- `MS365_MCP_TENANT_ID`: Custom tenant ID (defaults to 'common' for multi-tenant)
+- `MS365_MCP_TENANT_ID`: Custom tenant ID (defaults to 'common' for multi-tenant). **Personal Microsoft accounts should set this to `consumers`** - as of June 2026, refresh tokens issued via the default 'common' authority are rejected at the first refresh, so sessions die roughly an hour after login
 - `MS365_MCP_OAUTH_TOKEN`: Pre-existing OAuth token for Microsoft Graph API (BYOT method)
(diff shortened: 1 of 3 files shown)


eirikb:
Awesome work! Thanks :slightly_smiling_face: 

Do you think we should close #522? I honestly can't see any better solution than this...

koopatroopa:
No worries!

Yes, I'd close it. The consumers hint covers the practical fix, and anything fancier (auto-switching authority) is untested and better tracked separately if anyone ever needs it.

Thanks for the quick merges!

PR #533, opened 2026-06-15
feat: copilot retrieve endpoint

eirikb:
Hi there @lestephen

#501 was solid and the tool works, so this isn't me undoing it. More like wiring it into the rest of the machinery, and I'd really like your honest take on one part of it.

## What it does

Moves `copilot-retrieve` from a hand-rolled utility tool into a normal `endpoints.json` entry, so it behaves like every other Graph tool:

- shows up under `--preset search` / `--preset work` (before, it was invisible to any preset)
- requests its own scopes at login, and gets gated when they're missing
- work-only / org-mode gated, so it stops showing up on personal accounts where the API isn't supported anyway

The Graph request itself is byte-for-byte identical. Same `POST /copilot/retrieval`, same body.

## The part I want your opinion on

The call shape changes. You had nice flat params (`queryString`, `dataSource`, ...). Through `endpoints.json` it ends up nested under `body`:

```
copilot-retrieve({ body: { queryString, dataSource, ... } })
```

which is how every other POST tool here works. I know you chose flat on purpose, and honestly flat is the nicer shape on its own. My hunch is that consistency with the ~90 other tools wins for an LLM that already expects the `body` convention, buuut you probably know more than me about how it's actually being used. So does the nesting bother you? Any reason you'd rather keep it flat, or keep it a utility tool? I'd rather hear it now than guess 🙂

## The slightly hacky bit

I had to add a `requestBodySchema` field. Turns out `/copilot/retrieval` IS in Microsoft's OpenAPI spec, but only as a deprecated preview op with a malformed body our generator can't consume. So I declare the typed body in `endpoints.json` and let the generator build from that instead. Not pretty, but it keeps the enum + 1-25 / max-length validation you had.

There's a fair bit more reasoning behind all this (it also turned up a separate generator bug, fixed in #532). Can share the long version if you care, just didn't want to dump it all in here.

Could create a sidecar-json-file with spec - still cleaner than code (and perhaps cleaner than having it in endpoints.json, that part hurt me).

What do you think?


lestephen:
Thanks for looping me in @eirikb, and no worries reshaping #501. Moving it into the normal machinery makes sense; presets, login-time scopes, and personal-account gating are all things the hand-rolled tool couldn't get for free. Quick takes on your three, then a few things from a closer read of the diff.

**Nested `body`:** Don't keep it flat on my account. Consistency with the other ~90 POST tools is worth more than flat-on-its-own, and we control our own distribution so a shape change costs us nothing to follow, so please optimize for upstream.

**`requestBodySchema` inline vs sidecar:** I lean inline for now. It's opt-in and rare, and co-locating the schema with its endpoint beats a second file to keep in sync. Easy to lift into `requestBodySchemas/<toolName>.json` later if these accumulate. Your call on where the bloat threshold sits.

**Re #499:** worth noting `requestBodySchema` and the `apiVersion`/beta-routing from #499 are orthogonal: one is for "correct but beta-only" (Planner chat), the other for "published body unusable on any version" (`/copilot/retrieval`). So this doesn't preempt the generator work; it's a separate lever an endpoint could combine with beta routing later.

A few things I noticed reading the diff (none block us, since we run `--org-mode --preset all` with no scope allowlist, but flagging while they're cheap):

1. **`workScopes` is all-of, but these three scopes are dataSource alternatives.** `getEndpointRequiredScopes()` unions them and registration drops the tool if any is missing, so a tenant correctly permissioned for SharePoint/OneDrive (`Files.Read.All` + `Sites.Read.All`) or connectors (`ExternalItem.Read.All`) loses `copilot-retrieve` unless it over-consents to all three. Might be worth OR-groups (e.g. `[["Files.Read.All","Sites.Read.All"],["ExternalItem.Read.All"]]`) or splitting per `dataSource`, or just documenting that all three are required today. Whether to change the gate semantics is your call; there's a security dimension I won't pretend to weigh.

2. **The `requestBodySchema` branch replaces the whole operation, not just the body.** In `simplified-openapi.mjs` the `|| endpoint.requestBodySchema` path still does the wholesale `pathSpec[methodLower] = {…}` assignment, which discards MS's published responses/params for that op (and would mask any future upstream fields for any endpoint that opts in). Might be worth patching only `requestBody.content['application/json'].schema` and leaving the rest of the published operation intact, which keeps the override surgical and avoids silent drift if MS later fixes the spec.

3. **Read-only hint:** the generic registration derives `readOnlyHint`/`destructiveHint` from the HTTP verb, so this POST lands as `destructiveHint: true` even though it's a read-only query (the old utility set `readOnlyHint: true`). `endpointConfig.readOnly` already gates availability in read-only mode but isn't used for the emitted hint, so clients/policy layers that honor hints might mis-rank it.

Smaller: the deleted handler trimmed `queryString` (so `"   "` was rejected), but `minLength: 1` doesn't; a `pattern` would restore that. And the removed unit tests were the only coverage for the enum/range guards, so it might be worth one test through the generated path so a schema-translation regression doesn't go silent.

Happy to take a pass at any of these if it'd save you time. Since it's your branch, easiest is probably for me to open a PR against `feature/eirikb/copilot-retrieve-endpoint` (or hand you a patch / push a branch on my fork you can pull) for whichever ones are worth doing, say the OR-group scopes and the surgical `requestBody` merge. Just point me at the ones you want and I'll wire them up.

eirikb:
Awesome and thanks for the really good review.

I just did #2 and #3 on the branch, plus restored the whitespace guard with a `pattern` and added a test through the generated path.

Heads-up on #2: the surgical version made `copilot-retrieve` vanish, because MS ships `/copilot/retrieval` as `deprecated: true` and openapi-zod-client skips those by default. So I strip the flag too now. That's why the old wholesale replace "worked".

#1 I'll do as a separate PR - at login we request a concrete scope set, so "either/or" consent needs its own think. And yeah, the `llmTip` describes the API OR-semantics while the gate unions all three today, the OR-group PR reconciles that.

Thanks again :slightly_smiling_face:

PR #537, opened 2026-06-15
feat(generator): dual-generator support for Graph beta endpoints

eirikb:
Adds a per-endpoint "apiVersion" field in endpoints.json (default "v1.0"). Endpoints flagged "beta" are trimmed from Microsoft's parallel /beta OpenAPI spec into a second generated client (client-beta.ts) and routed to the /beta surface at runtime. v1.0 endpoints are unchanged.

Generator:
- download both the v1.0 and beta specs
- trimmer filters endpoints by apiVersion; one trimmed spec + client per version
- generate-mcp-tools now takes explicit (trimmedFile, clientFile) and formats its output with Prettier so generate output is prettier-stable

Runtime (version is config-driven; generated clients stay version-agnostic):
- graph-tools merges both clients into the registration loop and discovery registry, threads config.apiVersion into request options, and strips either /v1.0 or /beta from pagination nextLinks
- graph-client builds the URL from options.apiVersion (default v1.0)
- beta tool descriptions are auto-prefixed with [beta] in registration and discovery, so instability is visible regardless of the llmTip

Adds get-my-profile (GET /me/profile) as the first beta endpoint, beta-routing tests, and extends the endpoints/client orphan check to cover both clients.


eirikb:
Hi @lestephen

This adds beta endpoint support - it trims a second client from Microsoft's /beta spec and routes endpoints flagged `"apiVersion": "beta"` there. v1.0 stuff is untouched.

First one in is `get-my-profile` (GET /me/profile). If you want a beta endpoint added, just drop it in endpoints.json with the apiVersion flag.

I would love your feedback on the PR before we merge. Anything really.

lestephen:
Thanks @eirikb, this looks great, and it lines up well with what we discussed on #499. It reads clean: `apiVersion` defaults to v1.0 so existing endpoints are untouched, the trimmer/client split keeps the generated clients version-agnostic, and config-driven routing lets the runtime pick the surface per request. It also confirms the point from the #533 thread, that `apiVersion` and `requestBodySchema` compose without either knowing about the other. Once this lands we'll rebase #499 (Planner task chat) to just declare `apiVersion: beta` instead of carrying the beta plumbing ourselves. The nextLink change (anchored `^/(v1\.0|beta)` strip, re-applying the request's own version) looks right, including for beta pagination.

One thing worth hardening before the pattern spreads, especially since we're about to add beta endpoints on top of it: the dual-client seam has no guard against an alias landing in the wrong client or in both. `allEndpoints = [...api.endpoints, ...betaApi.endpoints]` is a plain concat, so if a toolName ever ends up in both generated clients (a future entry duplicated across versions, or a path Microsoft ships in both specs), `buildToolsRegistry`'s `registry.set(alias, ...)` silently keeps the last one and `registerGraphTools` registers the MCP name twice, with the wrong path/prefix/scopes quietly winning. Two cheap guards catch that whole class:

1. assert `allEndpoints` has no duplicate alias at module load, with a hard error on collision; and
2. in the endpoints.json validation test, check each entry's alias against the client matching its `apiVersion` (beta in `client-beta`, v1.0 in `client`, never both) rather than the merged union. Right now the union check still passes even if a beta endpoint were generated into the v1.0 client, which is the one seam this PR introduces.

A couple of other minor things you might want to pick up:

- The per-test `vi.mock('../generated/client-beta.js', ...)` is now duplicated across every test file, so a new test that forgets it imports the real generated client. A single vitest setup file would let tests inherit it, and save another sweep when a third version shows up.
- `const allEndpoints` sits between two import statements; moving it below the imports keeps `import/first` happy.

The dedup assert and the version-aware validation test are the two I'd prioritize. Either way, good to see this land.

PR #538, opened 2026-06-16
fix: don't silently drop OData params absent from generated client

steffen-heil-secforge:
## Summary

When a recognised OData param (`$filter`, `$skip`, `$top`, etc.) was passed at runtime but had no matching `paramDef` entry — because the generated client omitted it for that endpoint — the value was silently dropped after the path-param fallback chain without reaching `queryParams`.

**Fix**: add an `else if (isOdataParam)` branch that forwards the param to `queryParams` using the already-computed `fixedParamName`.

## Test plan

- [ ] All 304 existing tests pass (`npm test`)
- [ ] Call `list-mail-folder-messages` with `$skip=10` — verify different results vs `$skip=0`
- [ ] Verify other OData params (`$filter`, `$orderby`, etc.) continue to work as before

### DIFF
--- a/src/graph-tools.ts
+++ b/src/graph-tools.ts
@@ -444,6 +444,11 @@ async function executeGraphTool(
           .replace(`{${camelCaseParamName}}`, encodedValue)
           .replace(`:${camelCaseParamName}`, encodedValue);
         logger.info(`Path param fallback: replaced :${camelCaseParamName} with encoded value`);
+      } else if (isOdataParam) {
+        // Fallback: OData param recognised by name but absent from generated client's parameter
+        // list — forward it as a query param rather than silently dropping it.
+        queryParams[fixedParamName] = `${paramValue}`;
+        logger.info(`OData param fallback: forwarded ${fixedParamName}=${paramValue}`);
       }
     }
 


eirikb:
Thanks! This look good, but I wonder about one thing; injecting $skip of every single GET - is that smart?
Won't it hit endpoints which might not support it?
Have you tested against endpoints that might not support it - and what happens then? - Especially the "directory"-endpoints. Perhaps a skip $skip (pun) for those?

steffen-heil-secforge:
Good catch — and $skip the pun was appreciated 🙂

I checked: 48 GET endpoints include `$skip` in the generated client (from the OpenAPI spec), 99 do not — including `list-users` and other directory endpoints, which Graph explicitly does not support `$skip` on.

The unconditional injection was over-broad, so I've reverted it. The registration side is back to the original conditional check — `$skip` only appears in a tool's schema if the OpenAPI-derived client already declares it.

What remains is just the **execution-path fix**: a new `else if (isOdataParam)` branch so that any recognised OData param that arrives at runtime but is absent from the generated client's `paramDef` gets forwarded to `queryParams` rather than silently dropped. This closes the gap without polluting schemas for endpoints that don't support the parameter.

eirikb:

PR #544, opened 2026-06-18
feat(auth): Support path suffix on --public-url

dtennander:
Summary
---------
When hosting this service under a sub-path we need to expose the .well-known urls with specific URLs for this service/resource.

To my understanding this fix would be the common solution given this entry in the RFC: [3.1 Protected Resource Metadata Request](https://www.rfc-editor.org/rfc/rfc9728.html#name-protected-resource-metadata-). I hope you agree, otherwise please feel free to provide another approach we can look into to support this case.

Key changes:
-------------
This change makes the resource_metadata field of the WWW-Authorization response header include any path included in the --public-url parameter as a suffix to the well-known url.

### DIFF
--- a/src/lib/microsoft-auth.ts
+++ b/src/lib/microsoft-auth.ts
@@ -2,10 +2,41 @@ import { Request, Response, NextFunction } from 'express';
 import logger from '../logger.js';
 import { getCloudEndpoints, type CloudType } from '../cloud-config.js';
 
-function buildWwwAuthenticate(req: Request, error: string, description: string): string {
+/**
+ * Build the resource_metadata URL per RFC 9728 §3.1.
+ * (https://www.rfc-editor.org/rfc/rfc9728.html#name-protected-resource-metadata-)
+ *
+ * When the resource identifier has a path component that should be seen as the resource path.
+ * And thus be inserted in the end of the well-known URI.
+ *
+ *   https://example.com                   → https://example.com/.well-known/oauth-protected-resource
+ *   https://example.com/tenant/ms-365-mcp → https://example.com/.well-known/oauth-protected-resource/tenant/ms-365-mcp
+ *
+ * When publicUrl is absent the request Host
+ * header is used as the origin and no additional path component is appended.
+ */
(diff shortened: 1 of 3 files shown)


eirikb:
Nice, thanks! 🙂 One thing I'm not 100% sure about: the server only registers `/.well-known/oauth-protected-resource` (exact match), so the suffixed url this advertises would 404 on us unless a proxy (yours?) rewrites it. Were you relying on a proxy for that, or should we just register the suffixed route here too so it's self-contained?

dtennander:
We were relying on the proxy to do that, but you are right that we probably should accept both options here to be self-contained! I was also looking into the handler and saw this:
```js
      app.get('/.well-known/oauth-protected-resource', async (req, res) => {
        const protocol = req.secure ? 'https' : 'http';
        const requestOrigin = `${protocol}://${req.get('host')}`;
 ...
        res.json({
          resource: `${requestOrigin}/mcp`,
          authorization_servers: [browserBase],
          scopes_supported: scopes,
          bearer_methods_supported: ['header'],
          resource_documentation: browserBase,
        });

```
Do we want change ```resource: `${requestOrigin}/mcp`,``` to also use the public-url param?

dtennander:
I added two new commits (e5adbaf, edbb2e1) that handles the two things we discussed. If you don't agree with them we remove them.

My biggest concern is that I don't know the history for why `requestOrigin` was used in the first place for resource...

dtennander:
I found 23278183, and it seems that my use-case is a bit different then what is optimized for there. We want local clients (claude/opencode) to be able to connect as well over the public URL, and we have no problem with other services e.i openwebui to go though the load-balancer. But this could be a change of behavior you don't want to have

eirikb:
Thanks. Could you fix the formatting? :slightly_smiling_face:

dtennander:
Of course! Should be fixed now!

eirikb:
Failing tests. I usually run `npm run verify`

dtennander:
Now tests should all be green! Sorry about that!

eirikb:

PR #547, opened 2026-06-20
Fix tool discovery for create-calendar-event and sync package-lock

### LINKED ISSUE #543: create-calendar-event description leads with wrong Graph op, breaking semantic discovery
## Summary

`create-calendar-event`'s description opens with boilerplate that has nothing to do with creating events. In MCP clients that select tools by search over descriptions (Claude with deferred/lazy tool loading, for one), this makes the tool effectively undiscoverable: a query like "schedule a meeting" never retrieves it. The only queries that surface it are ones matching the irrelevant lead text.

The description currently leads with:

> Create one or more multi-value extended properties in a new or existing instance of a resource. The following user resources are supported: ...

That's the Graph `multiValueExtendedProperties` operation's blurb, not event creation -- it appears to be mis-sourced from the wrong Graph operation in the upstream spec. The actually-useful text (subject / start / end / attendees / recurrence / isOnlineMeeting) is present, but it sits in a `llmTip` appended *after* all the boilerplate, where it carries little weight in retrieval.

## Evidence

Three discovery probes against a client doing search-based tool selection. Same session, server healthy, returning a complete and valid schema.

| Query | `create-calendar-event` retrieved? |
|---|---|
| `schedule new meeting appointment add to calendar` (intent) | No |
| `create-calendar-event` (literal tool name) | No |
| `create multi-value extended properties event instance` (the boilerplate) | Yes |

Intent phrasing misses. The literal tool name misses. Only the irrelevant lead text hits. Retrieval is keyed on the description body, and this body describes the wrong operation.

For contrast, `delete-calendar-event` and `get-calendar-event` retrieve fine on intent queries. Same tip-appended pattern, but they happen to open with accurate sentences ("Removes the specified event...", "Get the properties and relationships..."). Correct lead, and they work.

## Cause

Search-based tool selection weights the whole description, and the opening sentence dominates. When the lead describes a different operation, the tool ranks as that operation. A correct `llmTip` stapled to the end can't drag it back, because `buildToolDescription` always appends the tip *after* the base description (`graph-tools.ts`):

```ts
let toolDescription = withApiVersionPrefix(
  tool.description || `Execute ${tool.method.toUpperCase()} request to ${tool.path}`,
  endpointConfig
);
if (endpointConfig?.llmTip) {
  toolDescription += `\n\n💡 TIP: ${endpointConfig.llmTip}`;
}
```

## Suggested fix

1. Fix the base description for `create-calendar-event` so it *opens* with what the tool does -- create/schedule a calendar event -- instead of the extended-properties boilerplate. The lead sentence is the whole ballgame for retrieval.
2. Since `llmTip` is appended after the base description by design, it can't repair a bad lead. A trailing tip barely moves selection.
3. Audit the other generated tools for the same mis-sourced lead. `create-draft-email` and `reply-to-group-thread` are prime suspects -- if they also open with Graph noise, they'll be just as undiscoverable.

### Proposed mechanism: a `descriptionOverride` field in endpoints.json

Today there's no way to fix the lead from data alone. The base description comes from the generated client's `tool.description` (sourced from the Graph OpenAPI spec), and `EndpointConfig` has no description field -- only `llmTip`, which appends. So the only data-only lever cannot reach the problem.

Adding an optional `descriptionOverride` to `EndpointConfig` solves it cleanly and keeps the fix in `endpoints.json` per tool, no spec regeneration required.

Interface:

```ts
  llmTip?: string;
  descriptionOverride?: string; // replaces the Graph-spec base description for this tool
```

Assembly site (`buildToolDescription`, ~line 1009) -- keep `||` so an empty spec description still falls through:

```ts
let toolDescription = withApiVersionPrefix(
  endpointConfig?.descriptionOverride || tool.description || `Execute ${tool.method.toUpperCase()} request to ${tool.path}`,
  endpointConfig
);
```

Mirror it in the second assembly site, `toResultEntry` (~line 1278), which builds descriptions for `search-tools` results, or the two paths will disagree:

```ts
description: withApiVersionPrefix(
  config?.descriptionOverride || tool.description || `${tool.method.toUpperCase()} ${tool.path}`,
  config
),
```

This also fixes both retrieval layers at once: the description string that external clients (Claude's deferred-tool search) index, and the server's own BM25 `search-tools` index. `llmTip` only ever helped the latter, and only within its `TIP_EXCERPT_TOKENS` cap.

Note this turns a data-only change into a small code change (interface + two assembly sites). If a data-only fix is required, the only alternative is regenerating the client with a corrected upstream description, since `llmTip` provably can't reach the lead.

## Scope note

The retrieval ranking itself is client-side, so this isn't strictly a server bug. But the description text is the lever the server controls, and fixing the lead sentence (or exposing `descriptionOverride`) is the highest-leverage way to make these tools selectable in any client that does search-based tool routing -- which is most of them now.

CamSoper:
This pull request introduces support for overriding the default tool descriptions with a new `descriptionOverride` field, allowing more accurate and relevant descriptions to be surfaced in tool registration, discovery, and schema generation. Several functions are updated to consistently use this override, and new tests are added to ensure improved discovery for calendar-related actions.

**Enhancements for Tool Description Overrides:**

* Added the `descriptionOverride` field to the `EndpointConfig` interface in `src/graph-tools.ts`, with documentation explaining its purpose and usage.
* Updated the `create-calendar-event` entry in `src/endpoints.json` to provide a more relevant, user-focused description via `descriptionOverride`.

**Consistent Use of Description Overrides:**

* Modified tool registration, discovery, and search index functions in `src/graph-tools.ts` to prioritize `descriptionOverride` over the default description when present. This affects tool descriptions, discovery indexing, and schema generation. [[1]](diffhunk://#diff-0be11dfcc21fdb6d7ffe5fed7332a1ff4631245880bfc61ef449ee65b52b712cL1049-R1054) [[2]](diffhunk://#diff-0be11dfcc21fdb6d7ffe5fed7332a1ff4631245880bfc61ef449ee65b52b712cL1192-R1200) [[3]](diffhunk://#diff-0be11dfcc21fdb6d7ffe5fed7332a1ff4631245880bfc61ef449ee65b52b712cL1318-R1327) [[4]](diffhunk://#diff-0be11dfcc21fdb6d7ffe5fed7332a1ff4631245880bfc61ef449ee65b52b712cL1405-R1418)
* Updated the `describeToolSchema` function in `src/lib/tool-schema.ts` to accept and use `descriptionOverride` for the tool description field. [[1]](diffhunk://#diff-b77791194a159930fde3b2e0b4e3bbee982eb557bef70123eef9339abe3a4a27L22-R23) [[2]](diffhunk://#diff-b77791194a159930fde3b2e0b4e3bbee982eb557bef70123eef9339abe3a4a27L55-R56)

**Testing Improvements:**

* Added new semantic search test cases in `test/discovery-search.test.ts` to verify that queries like "schedule a meeting" and "add appointment to calendar" correctly match the `create-calendar-event` tool, demonstrating the effectiveness of the description override.

Fixes #543

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13348,6 +13348,24 @@
         "node": ">= 12"
       }
     },
+    "node_modules/tsup/node_modules/yaml": {
+      "version": "2.9.0",
+      "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.9.0.tgz",
+      "integrity": "sha512-2AvhNX3mb8zd6Zy7INTtSpl1F15HW6Wnqj0srWlkKLcpYl/gMIMJiyuGq2KeI2YFxUPjdlB+3Lc10seMLtL4cA==",
+      "dev": true,
+      "license": "ISC",
+      "optional": true,
+      "peer": true,
+      "bin": {
+        "yaml": "bin.mjs"
+      },
+      "engines": {
+        "node": ">= 14.6"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/eemeli"
+      }
+    },
     "node_modules/tsx": {
       "version": "4.21.0",
       "resolved": "https://registry.npmjs.org/tsx/-/tsx-4.21.0.tgz",
@@ -13831,6 +13849,24 @@
         "yaml": {
           "optional": true
         }
+      }
+    },
+    "node_modules/vitest/node_modules/yaml": {
+      "version": "2.9.0",
+      "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.9.0.tgz",
+      "integrity": "sha512-2AvhNX3mb8zd6Zy7INTtSpl1F15HW6Wnqj0srWlkKLcpYl/gMIMJiyuGq2KeI2YFxUPjdlB+3Lc10seMLtL4cA==",
+      "dev": true,
+      "license": "ISC",
+      "optional": true,
+      "peer": true,
+      "bin": {
+        "yaml": "bin.mjs"
+      },
+      "engines": {
+        "node": ">= 14.6"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/eemeli"
       }
     },
     "node_modules/web-worker": {
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -359,6 +359,7 @@
     "toolName": "create-calendar-event",
     "presets": ["calendar", "outlook", "personal"],
     "scopes": ["Calendars.ReadWrite"],
+    "descriptionOverride": "Create (schedule) a new calendar event — a meeting or appointment — on the user's calendar. Set subject, start/end times, time zone, location, body, and attendees; supports online meetings and recurrence.",
     "llmTip": "CRITICAL: Do not try to guess the email address of the recipients. Use the list-users tool to find the email address of the recipients."
   },
   {
--- a/src/graph-tools.ts
+++ b/src/graph-tools.ts
@@ -43,6 +43,10 @@ interface EndpointConfig {
   supportsTimezone?: boolean;
   supportsExpandExtendedProperties?: boolean;
   llmTip?: string;
+  // Replaces the Microsoft-supplied base description everywhere it is surfaced (tool
+  // registration, BM25 discovery index, search-tools, get-tool-schema). Use when the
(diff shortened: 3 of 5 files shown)


eirikb:

PR #550, opened 2026-06-22
feat(mail): support using a local file as the email body

### LINKED ISSUE #549: Feature request: supply email body (and attachments) from a local file path
## Summary

When sending mail or creating drafts (`send-mail`, `create-draft-email`, the reply/forward tools, `add-mail-attachment`, and the shared-mailbox variants), the body content must be passed **inline as a string** in the request payload. There is no way to point a tool at a `.txt` / `.html` file on disk and have the server use that file's contents as the message body (or as an attachment).

It would be great to be able to do something like:

> "Send an email to X using `./newsletter.html` as the HTML body."

…and have the server read the file, rather than requiring the client/LLM to slurp the whole file into the tool arguments.

## Current behavior

The body is taken verbatim from the inbound tool argument and forwarded to Graph as the `message.body.content` (or `comment`) string — e.g. `graph-tools.ts` builds the request body straight from the passed parameter value. The only `readFileSync` usages in the server are for its own config (`endpoints.json`, `package.json`), never for tool inputs. So today the file-reading has to happen on the client side, and the entire body ends up inlined into the MCP tool-call arguments.

## Why this is useful

- Large HTML emails (newsletters, templated messages) blow up the token/context budget when inlined into tool arguments; reading from disk avoids round-tripping the whole body through the model.
- Common, ergonomic workflow: keep a templated `.html`/`.txt` body in a file and just reference it.
- Same idea applies to attachments — referencing a local file path is friendlier than base64-encoding `contentBytes` inline.

## Proposed solution

Add optional file-path parameters to the relevant mail tools, e.g.:

- `bodyFilePath` — read the file and use its contents as `message.body.content`. Infer `contentType` from the extension (`.html` → `html`, else `text`), with an explicit override still honored.
- (Optional, separate scope) `attachmentFilePath` / accept a path in `add-mail-attachment` — read + base64-encode the file into `contentBytes`, inferring the file name and content type.

## Important scope/security consideration

This only makes sense for the **stdio** deployment, where the server runs on the same machine as the user's files. For the **HTTP / OBO** server the filesystem belongs to the server host, not the caller, so a path parameter would be a footgun (and a potential local-file-read vector). The feature should be gated so it is only active in stdio mode (or behind an explicit opt-in flag), and disabled for HTTP/OBO transports.

## Alternatives

- Status quo: the client reads the file and inlines the content. Works, but costs context for large bodies and isn't discoverable.

---

I searched open and closed issues and didn't find an existing request for this — happy to be pointed at one if it's a duplicate.

romeupalos:
## Summary

Adds an optional `bodyFilePath` parameter to the mail tools so a caller can use a local `.html`/`.txt` file as the email body instead of inlining the whole body into the tool-call arguments. This is handy for templated/large HTML emails, where inlining the body wastes context.

Closes #549.

Applies to: `send-mail`, `create-draft-email`, `send-shared-mailbox-mail`, `create-shared-mailbox-draft`.

## How it works

- **Declarative, per-endpoint** — each supported entry in `endpoints.json` gets a `bodyFile` config (`contentTarget` / `contentTypeTarget`) describing *where* in the request body the file contents go. The message text lives at different paths per endpoint (`Message.body.content` for the `sendMail` action vs `body.content` for a draft message), so the target is data-driven rather than hard-coded.
- **Param injection** reuses the existing mechanism (the same path as `account` / `timezone`): when an endpoint declares `bodyFile`, an optional `bodyFilePath` and `bodyContentType` (`html`/`text`, inferred from the extension when omitted) are added to that tool's schema.
- **Pure helper** `src/lib/body-file.ts` (`inferContentType`, `setByPath`, `applyBodyFile`) does the merge; the file read + error handling stay in the tool handler so a read failure becomes a clean MCP error result.
- **Case-insensitive merge** — Graph's `sendMail` action wraps the message under the capital `Message` parameter, while a draft message uses lowercase `body`. `setByPath` matches existing keys case-insensitively, so the file content is merged *into* whatever wrapper the caller supplied (`Message` or `message`) rather than ever creating a duplicate key (which Graph rejects with a 500 "property with the same name already exists").

## Security / gating

Reading a path off disk only makes sense in **stdio** mode, where the server runs on the caller's own machine. The parameter is gated behind a new `allowFileBody` flag derived from `!options.http`, so it is **not registered at all in HTTP / OBO / trust-proxy deployments** (where the filesystem belongs to the server host, not the caller).

## Backward compatibility

`bodyFilePath` / `bodyContentType` are optional and additive — existing inline-body usage is unchanged (covered by a regression test asserting a byte-identical request when no file path is given).

## Tests

- `test/body-file.test.ts` — unit tests for the helper (content-type inference, no-clobber deep set, case-insensitive key reuse, inline override).
- `src/__tests__/graph-tools.test.ts` — integration tests: HTML/text inference, explicit content-type override, file-not-found → `isError` (no Graph call), inline override, the duplicate-`Message` collision regression, case-insensitive wrapper reuse, the no-`bodyFilePath` regression, and the HTTP-mode gating check.

Full suite green (`npm run verify`): generate → lint → format → build → 502 tests pass.

## Out of scope (possible follow-ups)

- reply/forward tools (their body is a plain `comment` string, no content type).
- attachments from a file path.
- discovery-mode registration (`registerDiscoveryTools`).

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -1,5 +1,8 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { z } from 'zod';
+import os from 'os';
+import path from 'path';
+import { mkdtempSync, writeFileSync, rmSync } from 'fs';
 
 /**
  * We test executeGraphTool logic by importing it indirectly through registerGraphTools.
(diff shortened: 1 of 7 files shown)


eirikb:
Thanks, this is nice work and it does work locally.

My main worry is it goes against how we handle bulky content everywhere else. `upload-file-content` and `add-mail-attachment` both take the content inline (base64 in the body) rather than a local path, and `download-bytes` / `returnDownloadUrl` pass things by reference. So the server never reads the caller's filesystem - and this PR would be the first place it does.

That's also why it's stdio-only: over a remote/http connector (claude.ai, cowork etc) the file isn't on the server, so `bodyFilePath` can't help there. But the goal (not wasting context on a big body) is just as real remotely.

So I think a path is the wrong primitive here, and we'd want something inline or fetchable to stay consistent with `upload-file-content`. Could we hold this and think about that a bit? What do you think?

romeupalos:
Absolutely.

Btw, thank you for the great work on this repo.

I implemented this because sending test HTML emails was consuming too many
tokens and taking long time for the agent to type byte by byte itself.

During implementation, Claude also mentioned issues when running
remotely. I'm not sure how to make this appropriate for remote execution,
though.

Thank you.
Romeu

eirikb:
I've been going back and forth on this and I don't think there's a clean fit. A path only works in stdio and it'd be the first place the server reads the callers disk, but doing it inline instead wouldn't actually save the tokens that were the whole point...

So I'll close it for now. Really appreciate the work though, and feel free to reopen if you have an idea I'm missing 🙂

PR #572, opened 2026-07-18
Add llmTips to update-todo-task and create-todo-task for tool discovery

CamSoper:
These two endpoints ship with only the bare Graph docstring ('Update the properties of a todoTask object.'), while every sibling To Do endpoint carries an llmTip. Since tool search ranks BM25 over name, description, and llmTip, natural queries like 'mark to do item complete' fail to surface update-todo-task at all, and MCP clients conclude the server has no task-update tool. 

This PR give both endpoints keyword-rich tips matching the style of their siblings. Tested and verified on my fork.

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -1100,14 +1100,16 @@
     "method": "post",
     "toolName": "create-todo-task",
     "presets": ["personal", "tasks"],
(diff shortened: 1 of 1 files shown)


eirikb:

Issue #272, opened 2026-03-16
Token never persisted after device code auth on Windows (work/school account)


riekards:
**Environment:** Windows, Node 24.13, global npm install, cmd /c wrapper, --org-mode
**What happens:** Device code is issued, Microsoft auth completes successfully (confirmation screen shown), but verify-login always returns "No valid token found" and list-accounts returns empty.

**What I tried:**

- npx and global install
- Custom token cache paths (ignored, folder never created)
- APPDATA env variable explicitly set
- keytar confirmed working
- --login CLI exits immediately without polling

Issue #124, opened 2025-08-27
Bug: Running the MCP Server in Alpine Container (Librechat) fails


schmichri:
Error Message: 
```
/app $ npx -y @softeria/ms-365-mcp-server  
node:internal/modules/cjs/loader:1651
  return process.dlopen(module, path.toNamespacedPath(filename));
                 ^

Error: Error relocating /home/node/.npm/_npx/813b81b976932cb5/node_modules/keytar/build/Release/keytar.node: unsupported relocation type 7
    at Module._extensions..node (node:internal/modules/cjs/loader:1651:18)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1096:12)
    at Module.require (node:internal/modules/cjs/loader:1298:19)
    at require (node:internal/modules/helpers:182:18)
    at Object.<anonymous> (/home/node/.npm/_npx/813b81b976932cb5/node_modules/keytar/lib/keytar.js:1:14)
    at Module._compile (node:internal/modules/cjs/loader:1529:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1613:10)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1096:12) {
  code: 'ERR_DLOPEN_FAILED'
}
```

Steps to reproduce: 

Create Dockerfile based on librechat.

```
# Base librechat images
ARG VERSION=v0.8.0-rc3
FROM ghcr.io/danny-avila/librechat:${VERSION}

USER root

# Install Libsecret dep
RUN  apk add libsecret libsecret-dev

USER node
RUN npx -y @softeria/ms-365-mcp-server
```

node version here is `v20.19.4`

PR #573, opened 2026-07-20
fix(auth): avoid keychain token cache by default

mcdmags:
## Summary
- remove the default keytar/OS credential-store token-cache path
- use file-based token storage by default for local MSAL cache and selected-account metadata
- update token-storage docs and tests to reflect that local auth must not trigger Keychain/Credential Manager/Secret Service prompts

## Verification
- npm run generate
- npm run build
- npm test -- token-cache-storage auth-storage cache-stamp auth-paths
- npm test
- npm run lint
- npm run format:check
- rg scan for runtime keytar/Keychain/Credential Manager/Secret Service usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,9 +609,9 @@ Environment variables:
 
 ## Token Storage
 
-Authentication tokens are stored using the OS credential store (via keytar) when available. If keytar is not installed or fails (common on headless Linux), the server falls back to file-based storage.
+Authentication tokens are stored in file-based storage by default. The server does not use the OS credential store, `keytar`, macOS Keychain, Windows Credential Manager, or Linux Secret Service, so a local auth flow must not trigger an operating-system password prompt just to read or write the MSAL cache.
 
-**Default fallback paths** are relative to the installed package directory. This means tokens can be lost when the package is reinstalled or updated via npm.
+**Default paths** are relative to the installed package directory. This means tokens can be lost when the package is reinstalled or updated via npm.
 
 To persist tokens across updates, set custom paths outside the package directory:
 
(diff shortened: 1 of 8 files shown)


eirikb:
Hi, thanks for the PR!

One thing though - this removes keytar completely (also from package.json), but the title says "by default", so I got a bit confused about the intent.

For mac/windows desktop users, and my own linux too, keytar does work and gives encrypted-at-rest storage, so I'd rather not drop it for them. Could we instead keep it but default to file, behind something like `MS365_MCP_USE_KEYTAR`? That kills the prompt for everyone by default, but leaves the option for people who want it.

What do you think?

eirikb:
Hi @mcdmags, sorry for the slow reply here...

I just merged #624 which rewrote `token-cache-storage.ts` quite a bit, so this one conflicts now. But I think it also makes your change a lot easier.

The reason I hesitated was that keytar gave encrypted-at-rest storage and I didn't want desktop users to lose that. That's not really true anymore - the cache file is AES-256-GCM either way now, and keytar only holds the 32 byte key. So defaulting to file means you get an encrypted cache plus a `0600` key file next to it, which is a much smaller step down than before.

It doesn't fix your actual problem though, since keytar still gets called on load and save. So the prompts are still there. But gating those few call sites behind something like `MS365_MCP_USE_KEYTAR` should be a pretty small diff on top of the new code, much smaller than what you have here.

Would you be up for rebasing and rescoping to that? Happy to take it over if you'd rather not, just say the word :+1:

eirikb:
Hi @mcdmags, sorry for leaving this hanging so long.

I've changed my mind a bit since last time. keytar has actually been behind five separate issues here (#602, #418, #298, #272, #124), so having a way out of it makes sense. But I think I'll keep it on by default and add `MS365_MCP_USE_KEYTAR` to switch it off, rather than flipping the default. Flipping would leave anyone with an existing key in the keychain unable to decrypt their cache, and `assertOverwritable` would then refuse to write a new one, so they'd be stuck until they deleted the file by hand.

\#624 rewrote most of that file anyway so this conflicts quite badly now. Hope it is ok I'm closing it and doing the smaller version myself. Your idea though, thanks for pushing on it 🙂

PR #587, opened 2026-07-22
Add direct mail attachment downloads

cduguet:
## Summary

- add a `download-mail-attachment` utility tool that streams Outlook attachment bytes directly to the MCP server filesystem
- confine writes to `MS365_MCP_DOWNLOAD_DIR`, sanitize filenames, use exclusive mode-0600 files, and remove partial downloads
- return path, content type, and size metadata without sending base64 attachment data through the MCP response
- document the new configuration and update attachment guidance

## Why

Microsoft Graph mail attachments do not expose pre-authenticated download URLs. The existing `download-bytes` tool returns base64 through the MCP context; this tool lets local and hosted MCP servers persist authenticated attachment bytes directly instead.

## Validation

- `npm run build`
- `npm run lint` (0 errors; existing warnings remain)
- `npm run format:check`
- `npm test` (606 tests)
- real Outlook attachment smoke test: downloaded a 36,203-byte, two-page PDF with a valid `%PDF-` signature; the MCP response contained no attachment bytes

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,6 +587,7 @@ Environment variables:
 - `MS365_MCP_FORCE_WORK_SCOPES=true|1`: Backwards compatibility for MS365_MCP_ORG_MODE
 - `MS365_MCP_OUTPUT_FORMAT=toon`: Enable TOON output format (alternative to --toon flag)
 - `MS365_MCP_MAX_TOP=<n>`: Hard cap for Graph `$top` / `top` on list requests (positive integer). When the model passes a larger value, the server clamps it to `n` so responses stay smaller. Example: `MS365_MCP_MAX_TOP=15`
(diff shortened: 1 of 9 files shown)


eirikb:
Hi

Why is this in draft?
I'm not sure if it's coincidence but #589 solves the very same thing, albeit a bit differently.
Since this is in draft, would you mind if I accept the other one but cherry pick the parts from this PR I like?
I think a merged version of the two PRs might be a good idea.

eirikb:
Thanks a lot for this, and sorry for the parallel work - your streaming `downloadToFile` is basically what ended up in #591, so the approach lived on 🙂

The attachment-to-disk case is now covered by `download-bytes-to-file` (#589) plus the streaming in #591. The one bit I didn't pull over is your `MS365_MCP_DOWNLOAD_DIR` confinement - I kept the tool stdio-only, since writing files onto a remote host in HTTP mode felt like something most hosted users wouldn't really want. Buuut if you reckon the HTTP-mode download case is worth having, ping me and I'll look at it as its own thing.

Since this is a draft and the same thing has landed, I'll close it now. Feel free to reopen if you disagree.

PR #589, opened 2026-07-22
feat: add download-bytes-to-file — save Graph content to disk without base64 round-tripping

jakeswenson:
## Why

`download-bytes` returns binary content as **base64 through the agent context**. For anything but the smallest files that is a huge waste of tokens: the model receives a giant base64 blob and then has to *dutifully re-emit those same bytes* into a file-write tool call — paying for the payload twice (once in, once out) and often blowing the context window entirely on a single attachment.

For OneDrive/SharePoint files there's already an escape hatch (`get-download-url`, a pre-authenticated out-of-band URL). But Microsoft Graph exposes **no** such URL for **mail attachments** or **meeting recordings** — those bytes are only available through authenticated `$value` endpoints. So today the *only* way to save a large mail attachment is the base64-through-the-LLM path, which doesn't scale.

This tool closes that gap: the server reads the authenticated bytes and writes them straight to disk, so **zero file bytes pass through the model**. This matters a lot for real inboxes with multi-MB attachments and recordings.

## What

New stdio-only utility tool `download-bytes-to-file`:

```
download-bytes-to-file
  target:     "/me/messages/{id}/attachments/{id}/$value"   (same targets as download-bytes)
  outputPath: "/absolute/path/on/disk.pdf"                   (required, absolute)
  account?:   "..."                                          (multi-account only)
→ { path, contentType, bytesWritten }
```

- Same target surface as `download-bytes` (drive files, mail attachments, profile photos, Teams hosted content, meeting recordings).
- Returns only metadata — the bytes never enter the model context.

## Behavior & safety

- **stdio only.** Gated behind a new `httpMode` flag threaded through `registerGraphTools`/`registerDiscoveryTools`; hidden in HTTP/OAuth mode so a remote client can't write arbitrary files onto the host.
- **`readOnlyHint: true`.** `--read-only` in this codebase means "no Graph API writes" (it's a `GET`); the filesystem-write risk is guarded by `stdioOnly`, not the read-only flag — so it stays available under `--read-only`.
- **Absolute `outputPath` required**, and it **never overwrites** an existing file (fail-fast `access` check + atomic `wx` write flag closing the TOCTOU gap).
- Reads the bytes via the low-level `makeRequest` (raw `{ encoding, contentBytes }` / `{ rawResponse }`) rather than the `graphRequest` MCP-formatting wrapper. This avoids a needless serialize→parse round-trip of a large base64 blob, works correctly under `--toon` (where the wrapper would be TOON, not JSON), and surfaces real Graph HTTP errors (404/403/429/…) instead of masking them.

## Testing

- New tests cover: binary write + decode, text/JSON body write, relative-`outputPath` rejection, absolute-URL `target` rejection, overwrite refusal, Graph-error passthrough (no file written), and stdio-vs-HTTP mode gating.
- Updated utility-tool count/args assertions in `tool-filtering`, `read-only`, and `allowed-scopes` tests.
- Tuned discovery `searchKeywords`/description so `get-download-url` keeps ranking first for drive/SharePoint download queries.
- Full suite green (611 tests).


eirikb:
Hi @jakeswenson - really nice work, this is a clean base. Also something I've noticed myself but not gotten around to do. Merging it 🙂

I'll layer a couple of small follow-ups on top - wiring it into the tool presets, and maybe borrowing the streaming bit from #587 for the big files. Nothing needed from you. (you might have seen I mentioned this PR in that PR...)

jakeswenson:
Hey @eirikb yeah, actually i was going to submit another pr where there is a new `download` preset that has this tool... honestly i don't think `download-bytes` should be ever used since i had a lot of trouble with it and my agent spun/looped with it quite a bit...

PR #594, opened 2026-07-26
fix(mail): correct six /me/messages tool descriptions mis-sourced from the Graph spec

msoukhomlinov:
Microsoft's OpenAPI metadata attaches the wrong description to every
/me/messages operation. The operationIds are right, but the prose is
copied from unrelated resources, so six mail tools describe something
they do not do:

  list-mail-messages     openTypeExtension (get)
  create-draft-email     openTypeExtension (create)
  get-mail-message-mime  openTypeExtension (get)
  get-mail-message       eventMessage
  update-mail-message    eventMessage
  delete-mail-message    eventMessage

list-mail-messages, for example, opens with "Get an open extension
(openTypeExtension object) identified by name or fully qualified name."
There is no signal in the spec a generator could use to detect this, so
descriptionOverride is the only fix - same approach as #547 for
create-calendar-event in #543.

This matters for clients that rank tools by description. Where a
description is truncated for display, the model can see nothing but text
about a different API.

Adds six golden queries to the discovery search test.

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -10,6 +10,7 @@
     "pathPattern": "/me/messages",
     "method": "get",
     "toolName": "list-mail-messages",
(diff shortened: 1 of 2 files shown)


eirikb:
Thank you, and great find! :slightly_smiling_face: 

Not the first time we've seen the spec can't be trusted, which hurts LLMs.

PR #595, opened 2026-07-26
feat(schema): describe $expand as navigation-properties-only

msoukhomlinov:
Every $expand in the generated client carries Microsoft's
"Expand related entities", which says nothing about what is actually
expandable. Models pass non-navigation properties - message body is the
one I hit repeatedly - and Graph answers

  400 Parsing OData Select and Expand failed: Property 'body' on type
  'microsoft.graph.message' is not a navigation property or complex
  property. Only navigation properties can be expanded.

$expand is the one OData parameter the describe() overrides in
registerGraphTools do not cover; $filter, $search, $select, $orderby,
$top, $skip and $count all get useful text. This adds $expand to that
chain, states the constraint, and points at $select for ordinary fields.

Kept deliberately generic. An earlier draft listed common navigation
properties per resource, but the override applies to all 140 $expand
params, most of them on drives, workbooks, groups and chats, where
mail and calendar examples would be noise. It also named several
properties I could not confirm are expandable in v1.0.

The schema is restated rather than changed: $expand is already
array<string> on every parameter it applies to.

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -64,6 +64,7 @@ function makeEndpoint(overrides: Partial<any> = {}) {
       { name: 'search', type: 'Query', schema: z.string().optional() },
       { name: 'select', type: 'Query', schema: z.string().optional() },
       { name: 'orderby', type: 'Query', schema: z.string().optional() },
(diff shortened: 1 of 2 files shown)


eirikb:
Nice, thanks!

PR #596, opened 2026-07-26
fix(schema): tell the model which path-parameter name to use

msoukhomlinov:
Path parameters reach the model with no usable description. Microsoft's
metadata leaves them out of most operations' parameter lists, so
generated/hack.ts synthesizes them as

  Path parameter: messageId

and the operations that do declare them describe them as

  Path parameter: drive-id

Neither says where the value comes from, and the second is spelled
differently from the name a caller actually has to send. Graph listings
return the identifier under the generic key `id`, so models pass `id`,
the parameter is rejected as missing, and retrying does not help because
nothing in the schema hints at the real name.

Both cases now get:

  Value for the 'messageId' path segment. Pass it under the name
  'messageId', not as 'id'. Use the 'id' field of the message object as
  returned by Microsoft Graph.

Deliberately narrow about what it claims:

- Non-identifiers get only `Value for the 'address' path segment.` A
  fallback that pointed at "the corresponding list tool" would be wrong
  for get-excel-range.address (an A1 range), search-onedrive-files.q
  (search text), list-supported-time-zones.TimeZoneStandard (an enum)
  and get-sharepoint-site-by-path.path.
- It says "as returned by Microsoft Graph" rather than naming a list
  tool, because placeId, plannerPlanId and virtualEventWebinarId have
  none.
- A numeric suffix gets no resource name at all. In
  /messages/:chatMessageId/replies/:chatMessageId1 the numbered
  parameter is the reply, not the parent.

The text lives in one helper so the two sites cannot drift. Tests assert
against the real generated client and against literal expected strings,
not against the helper's own output: hack.ts supplies parameters before
registerGraphTools sees them, so a mock-only test passes while
production is unchanged.

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -532,6 +532,25 @@ describe('graph-tools', () => {
       // Names at least one real navigation property so the model has something to copy.
       expect(schema['expand'].description).toContain('attachments');
     });
+
+    // graph-tools synthesizes path params only for endpoints where the generated
+    // client (via hack.ts) has not already supplied one — mostly function-style paths.
+    it('should describe path params it synthesizes itself', async () => {
+      const endpoint = makeEndpoint({ path: '/me/messages/:messageId' });
+      const config = makeConfig();
+      mockEndpoints.push(endpoint);
+      mockEndpointsJson = [config];
+
+      const server = createMockServer();
+      const { registerGraphTools } = await loadModule();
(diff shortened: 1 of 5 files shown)


eirikb:
Hi, good catch, and thanks. Pulled it down, merges clean and tests pass, so I'll merge this.

From a quick local test it seems to add around 7k tokens to the tool list, which is a lot but I think is worth it, and anyone who cares can run with `--discovery` anyway.

PR #597, opened 2026-07-26
fix(mail): normalize $search KQL quoting instead of returning 400

msoukhomlinov:
Outlook mail wants the whole KQL expression inside one pair of double
quotes. Models get this wrong in two directions, and both are 400s:

  "from:john" AND subject:meeting     quotes each clause
  subject:"quarterly report"          phrase, no enclosing pair

The llmTips already say to wrap the whole expression and it still
happens, because the schema line says "wrap value in double quotes",
which reads as "each value".

Checked each form against Graph before picking the rule:

  flooring                                   200
  "subject:flooring"                         200
  "subject:\"quarterly report\""             200
  subject:"quarterly report"                 400
  "quarterly report" AND from:x              400

So the fix is to always produce one enclosing pair, and quotes inside it
have to be read rather than stripped. A quoted run is either a phrase,
whose quotes group the words, or a whole clause the caller quoted by
mistake. Two signals separate them: a run introduced by `property:` is
always a phrase, even when its own text contains a colon
(subject:"RE: quarterly report"); otherwise a run that itself starts
with `property:` is the mistake. Phrase quotes are kept, escaped as \" —
the form Microsoft documents for embedded quotes.

A clause is recognised by a known message property followed by `:` or a
comparison — `from:john`, `size>1000`, `received>=2024-01-01`, the date
form endpoints.json advertises. Shape alone is not enough: "RE: quarterly
report" and "Q3: plan.pdf" both look like property:value, and dropping
their quotes still returns 200 while searching different terms. An
unrecognised property is therefore left alone rather than repaired —
leaving a query unfixed is recoverable, changing what it means quietly is
not.

An expression already inside one enclosing pair is unwrapped first so its
interior is judged on its own terms, otherwise correctly wrapped free
text ("quarterly report") would be read as a phrase and narrowed to an
exact-phrase search. An unbalanced quote keeps its text and loses the
stray delimiter. A value that is only quotes or whitespace is dropped
rather than sent.

Scoped to Outlook mail: the path must be under /me or /users AND name a
messages or mailFolders collection. Everything else is left untouched,
because the grammars differ and a blind rewrite would corrupt rather
than repair:

- directory search documents clause-level quoting,
  $search="displayName:john" OR "displayName:jane"
- /chats, /teams and /planner have their own /messages collections
- /sites and list-relevant-people document unquoted free text

Idempotent: normalizing an already-correct expression returns it
unchanged. Runs beside the existing clampTopQueryParam and logs like the
Body auto-wrap in executeGraphTool.


eirikb:
Hi

Nice one, and thanks for the digging in the description. Checked the docs and the mail vs directory split looks right.

Two small things before I merge:

* `SEARCH_PARAM_DESCRIPTION` in `src/lib/param-descriptions.ts` still says "wrap value in double quotes" - that's the exact wording you point at as the cause, could you fix that line too?
* When the value is unsearchable you delete `$search`. I think that turns a search into a full inbox listing, which is worse than the 400. Rather error out?

Also, do you still have the raw response for `subject:\"quarterly report\"`? The escaping isn't in the docs for messages, only for directory objects, and the tests mock Graph so I can't really check it.

Also, conflicts now (that's on me for being slow to reply, but still need to resolve).

eirikb:
Hi @msoukhomlinov

Sorry for the slow reply. Merged main in and pushed some commits on top, hope that is ok.

Both your points are in. Empty `$search` errors now instead of deleting the param. The description is fixed too, buuut I made it shorter rather than longer, it hits around 100 tools and only a handful are mail.

That `\"` escaping I asked about, I cannot find it documented for messages at all, only for directory objects. Left it as is but marked it inferred.

Then I got carried away. Some review agents pointed out that `"subject:quarterly report" AND from:john` lost the phrase, and that `from:"a\\"` came out unbalanced. Patched those one at a time and broke something new each round, so I ended up rewriting the classification instead. Also narrowed the path check, it was catching `/me/mailFolders` and `list-mail-rules`.

Verify is green, 1211 tests. Shout if you disagree with any of it, I have been quite heavy handed with your PR here.

PR #599, opened 2026-07-26
fix(discovery): share param-description overrides between registration paths

msoukhomlinov:
## Summary

- `--discovery` mode's `get-tool-schema` (`describeToolSchema` in `src/lib/tool-schema.ts`) built parameter descriptions straight from the raw, generated-client parameter list, while normal tool registration (`registerGraphTools` in `src/graph-tools.ts`) rewrites several of the same parameters into spec-gap-guidance text every time a tool is registered. The two paths built the same conceptual thing independently, so `describeToolSchema` was frozen at the raw Microsoft spec text and never picked up any override.
- Codex flagged this for `$expand` on an unrelated PR: Microsoft's spec text ("Expand related entities") tells a model nothing, and a merged fix rewrote it — but only in `registerGraphTools`. `--discovery` mode kept showing the useless text regardless. Auditing the rest of the override block found the same risk on every OData parameter it rewrites (`$filter`, `$search`, `$select`, `$orderby`, `$top`, `$skip`, `$count`), plus several synthetic parameters `describeToolSchema` never surfaced at all (`account`, `timezone`, `expandExtendedProperties`, `fetchAllPages`). `$top` also needed the `TOP_UNSUPPORTED_DELTA_TOOLS` exclusion (registerGraphTools deletes it entirely for two delta tools that don't support it) replicated as an omission, not just a re-description.
- Fix: extracted every override's text plus the associated conditions into a new `src/lib/param-descriptions.ts`, consumed by both `registerGraphTools` and `describeToolSchema` so they cannot independently drift again. Also absorbed the `confirm` param's description (previously two independently-maintained copies of identical text) into the same shared source.
- Verified `describeUtilityToolSchema` (download-bytes, get-download-url, ...) is unaffected — it calls `utility.buildSchema(ctx)` directly, the same closure the non-discovery path already uses.

## Test plan

- [x] `npm ci`
- [x] `npm run generate` (regenerated the Graph client from live Microsoft specs; re-ran build+tests after)
- [x] `npm run build`
- [x] `npm run format:check`
- [x] `npx eslint . --quiet`
- [x] `npx vitest run` — 643/643 tests pass across 51 files
- [x] New tests in `test/tool-schema.test.ts` assert `describeToolSchema`'s output matches, byte-for-byte, what `registerGraphTools` actually registers (via a real `McpServer` spy) for `$expand`, `$select`, `$filter`/`$search`/`$orderby`/`$skip`/`$count`, `$top` (including the delta-tool exclusion), `timezone`, `expandExtendedProperties`, `fetchAllPages`, `confirm`, and `account`
- [x] Manual end-to-end proof: built `dist/index.js`, ran it for real with `--discovery` over stdio via the MCP SDK client, called the real `get-tool-schema` tool:
  - `get-mail-message` → `$expand` now returns the fixed navigation-properties guidance (previously would have been the raw "Expand related entities")
  - `list-calendar-events-delta` → `$top`/`top` correctly absent from the parameter list
  - `get-calendar-view` → `timezone` and `expandExtendedProperties` now present with the same text `registerGraphTools` uses

### DIFF
--- a/src/graph-tools.ts
+++ b/src/graph-tools.ts
@@ -30,6 +30,28 @@ export interface DiscoverySearchIndex {
   nameTokens: Map<string, Set<string>>;
 }
 import { describeToolSchema, describeUtilityToolSchema } from './lib/tool-schema.js';
+import {
+  TOP_UNSUPPORTED_DELTA_TOOLS,
+  shouldOmitTopParam,
+  paginationAllowed,
+  positiveIntFromEnv,
+  DEFAULT_MAX_PAGES,
+  getMaxPages,
+  isFetchAllPagesApplicable,
+  FILTER_PARAM_DESCRIPTION,
+  SEARCH_PARAM_DESCRIPTION,
+  SELECT_PARAM_DESCRIPTION,
+  EXPAND_PARAM_DESCRIPTION,
+  ORDERBY_PARAM_DESCRIPTION,
+  TOP_PARAM_DESCRIPTION,
+  SKIP_PARAM_DESCRIPTION,
+  COUNT_PARAM_DESCRIPTION,
+  CONFIRM_PARAM_DESCRIPTION,
+  TIMEZONE_PARAM_DESCRIPTION,
+  EXPAND_EXTENDED_PROPERTIES_PARAM_DESCRIPTION,
+  getAccountParamDescription,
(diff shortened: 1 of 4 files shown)


eirikb:

Issue #168, opened 2025-11-28
PDF attachments retrieval issue


imoraru-nf:
Hello.. this is a great product, but i seem to have an issue with the PDF attachments :(
When they are fecthced and converted to base64, they are returned truncated (or at least this is what Claude pretends it is happening when it tries to retrieve them )

```
BINGO! YOU FOUND IT! 🎯🎯🎯
The Smoking Gun:

n8n workflow base64: 46,968 characters (COMPLETE)
MS365 MCP gave us: 4,561 characters (TRUNCATED!)

The MS365 MCP server is only giving us approximately 10% of the actual PDF data!
The base64 string is being severely truncated - it cuts off at character ~4,561 when it should be 46,968 characters long.
```

eirikb:
Mail attachments?

eirikb:
Alright, thanks, I'll take a quick look when I get some free time

imoraru-nf:
i attache here the chat containing the base64 conversion that was returned (truncated)  .. but also the base64 conversion returned by n8n workflow when fetchign the same attachment (non-truncated) ...maybe it helps troubleshooting :)

[Claude chat (inluding returned base64 string).txt](https://github.com/user-attachments/files/23818886/Claude.chat.inluding.returned.base64.string.txt)

[1_base64_full.txt](https://github.com/user-attachments/files/23818893/1_base64_full.txt)

rprouse:
I'm also running into this. I'll see if I can carve out some time to reproduce in a dev environment and provide more information and/or a fix.

eirikb:
Hi. I've been working on a big file handling fix. Will hopefully push it tomorrow

eirikb:
Hopefully solved by #439

Issue #574, opened 2026-07-20
feat: Extraction of File Content


michimalek-besitec:
I love the MCP, we deployed it for internal Enterprise Access to our M365 Ecosystem. We tested it thoroughly and enjoy the MCP a lot. We were wondering, is there a reason the MCP is missing and file/content-extraction features? You can gather files and their URL, but if you require the file or content in-chat, this is not a possibility to my current knowledge. Am I missing something or is this a planned feature in the backlog?

I created a second independent MCP-server that can actually be used combined with the ms365 MCP that allows the extraction of files using OCR and other methods.

Very curious!

eirikb:
Hi there, and thanks for the kind words 🙂 

It should be possible to pull in content already, using some utility tools we have:

  * downloads-bytes (base64), and
  * get-download-url gives a pre-authed download URL for bigger files to run out-of-bands of LLM (straight to disk)

For text-ish files (txt, md, csv, json, etc.) should decode to readable content directly. But for PDFs, docx, scanned images etc the raw bytes aren't useful text on their own, and that's the part that's deliberately missing: Microsoft Graph has no OCR or text-extraction API for me to wrap. So pairing it with a dedicated extraction/OCR MCP like you did is exactly how I'd do it too, not really a workaround.

We had another PR earlier that had a solution for this, but it wasn't really OCR and it did a bit more than I had in mind.

Small bonus, according to claude: Graph can convert Office docs to PDF and that passes straight through, e.g. target `/drives/{drive-id}/items/{item-id}/content?format=pdf`.

eirikb:
Closing this now, feel free to re-open if you want to.

PR #373, opened 2026-04-10
feat: add text extraction tool for OneDrive and SharePoint files

jmleksan:
### Summary

Adds **`download-drive-file-text`**: download a OneDrive / SharePoint drive file via Microsoft Graph, extract readable text (PDF, DOCX, plain text and similar), return structured JSON. Introduces a small **server tool extensions** layer so non–OpenAPI tools participate in **discovery mode** (`search-tools` / `execute-tool`) with stable synthetic `method` + `path` metadata, and refactors **`parse-teams-url`** into that same pattern.

### Motivation

- Assistants need full document text from Graph-backed files, not only a download URL.
- Utility tools that are not generated from OpenAPI should still be **discoverable and executable** in discovery-only deployments, with copy/paste-friendly “API-like” descriptors for `search-tools`.

#### Server tool extensions (`src/tool-extensions/`)

- Goal is to make it simple and organized for adding other tools that are not standard api calls like the team's tool and this new tool
- New **`SERVER_TOOL_EXTENSIONS`** registry (`index.ts`): each entry has `name`, synthetic **`method` / `path`**, `description`, **`execute`**, and optional **`buildMcpRegistration`** for normal MCP mode.
- **`parse-teams-url`**: moved from inline `graph-tools.ts` into `parse-teams-url.ts`; registered as extension with `POST` + `/utility/parse-teams-url`.
- **`download-drive-file-text`**: new extension with `GET`-style path description reflecting Graph download + extraction (`/drives/{drive-id}/items/{driveItem-id} (download bytes, extract text)`).
- **`registerServerToolExtensionsAsMcpTools`**: registers top-level MCP tools when `buildMcpRegistration` is set; respects **`--enabled-tools`** regex like Graph tools. Discovery still lists **all** extensions.
- **`types.ts`**: shared `McpRegistrationContext` / `McpToolRegistration` for Zod schemas and hints.

#### Discovery mode (`src/graph-tools.ts`)

- **`search-tools`**: includes extension tools in results; **`total`** and copy reflect Graph count + extension count; tip clarifies **non-OpenAPI extensions use the same `execute-tool` interface**.
- **`execute-tool`**: resolves **`tool_name`** against the Graph registry first, then against **`SERVER_TOOL_EXTENSIONS`**.
- Inline `parse-teams-url` registration removed in favor of the extension registrar.

#### Graph client (`src/graph-client.ts`)

- **`GraphRequestOptions`** exported for callers (e.g. drive download with token override).
[...]

PR #603, opened 2026-07-27
feat: add convert-document, a server-side attachment/file-to-markdown tool

msoukhomlinov:
Fetching an email attachment or drive file today means `download-bytes`
returning base64 that a model has to hold in its own context and cannot
parse anyway - a 73KB PDF is ~24,500 tokens of unusable bytes. This adds
a tool that fetches the same Graph binary content and converts it to
markdown on the server, using the optional `officeparser` package
(PDF, DOCX, PPTX, XLSX, ODT, RTF and a few other formats). On a real
3-page quote PDF this took the same request from ~24,500 unusable tokens
to ~2,100 tokens of the actual pricing table.

This is the "special endpoint on the MCP server which is not just a
direct proxy to MS Graph" floated in #168, and the token-cost problem is
the same one #549 was closed over - that one couldn't find a clean fit
because a caller-supplied disk path only works in stdio and would have
been the first place the server reads the caller's filesystem.
`convert-document` never touches a path in either direction: it fetches
bytes over the network exactly like `download-bytes` already does, and
returns text. No new filesystem surface, so it works over HTTP.

## What it adds

- `convert-document` in `graph-tools.ts`'s `UTILITY_TOOLS`, built the same
  way as `download-bytes` (same target semantics, same account handling)
  but decoding `contentBytes` and converting instead of returning them.
  A 25MB source-size cap, checked against the actual decoded buffer (not
  a reported `contentLength`), runs before conversion is attempted.
- `--enable-document-conversion` (off by default): this is untrusted
  binary content going through a new dependency, so it's opt-in rather
  than on by default like `download-bytes`. Folded into every preset's
  combined pattern in `tool-categories.ts` the same way `download-bytes`
  and `download-bytes-to-file` are, since it reads the same universe of
  binary paths they do.
- `officeparser` added to `optionalDependencies`, mirroring the existing
  `@azure/identity` / `keytar` precedent for capabilities most installs
  don't need. Missing-dependency and conversion-failure paths both
  return a normal tool error rather than crashing the server.
- A Node `>=22.13.0` startup guard specific to this feature (officeparser's
  own dependencies - `file-type`, `pdfjs-dist` - require it, even though
  officeparser itself declares broader support and the package as a whole
  still supports Node >=18).

## Why a worker thread

The parse runs in a real `worker_threads` worker with a hard timeout
enforced by *terminating* that worker, not by asking `officeparser` to
cooperate. An `AbortSignal`-based timeout next to the parse call was
tried first and doesn't work for every format: officeparser's Excel/Word/
PowerPoint parsers only check for cancellation between loop iterations
with no `await` inside them, so on a large-but-ordinary spreadsheet
(nothing pathological - just wide) the whole single-threaded server sat
blocked for the entire parse regardless of the configured timeout, since
the timer callback that would trigger the abort never got a turn on the
event loop. `worker.terminate()` stops the parse regardless of whether
its own code ever checks anything, and the worker's heap is separately
capped (`maxOldGenerationSizeMb`, default 512MB), so a pathological
document exhausts only its own worker's budget.

## Hardening since the first draft

This went through several rounds of review (a self-run adversarial QA
pass plus Codex, both against the fork branch) before this PR. Folding
that in now rather than layering it on after merge:

- Markdown truncation happens *inside* the worker, before the result
  crosses the thread boundary - otherwise structured cloning would copy
  the full untruncated string onto the main heap before the
  20,000-character cap is applied, defeating the point of the cap on a
  document that expands a lot (e.g. a small compressed XLSX with one
  very long cell).
- A process-wide concurrency limit (`MS365_MCP_MAX_CONCURRENT_CONVERSIONS`,
  default 3) is reserved *before* the Graph fetch, not just around the
  worker - otherwise several concurrent calls could each buffer up to the
  25MB source cap before any of them hit the concurrency check.
- The 25MB source-size cap applies to *both* the binary and non-binary
  response branches in `graph-client.ts` - a large `text/plain` or (at
  the time) `application/rtf` target took the unguarded `response.text()`
  path instead, bypassing the cap entirely.
- A Graph response can stall or trickle mid-body after headers arrive;
  `fetchWithResilience`'s own timeout is cleared the moment `fetch()`
  returns, so a per-chunk stall deadline (reusing the same
  `MS365_MCP_GRAPH_TIMEOUT_MS` config) was added to `readBodyWithLimit`
  so a stalled connection can't hold a reserved conversion slot forever.
- `isBinaryContentType` now also classifies legacy Word (`application/msword`)
  and RTF (`application/rtf`, `text/rtf`) as binary - both are formats
  officeparser explicitly supports, but neither matches the existing
  `vnd.`/`x-` vendor-prefix rule, so Graph responses for them were being
  UTF-8-decoded (corrupting the bytes) instead of preserved.
- Every worker settlement path (timeout, reported success/failure, and an
  uncaught worker error) now awaits `worker.terminate()` before resolving
  or rejecting, rather than firing termination and settling immediately -
  otherwise a caller releasing its concurrency slot in a `finally` could
  do so before a large worker's heap had actually been freed.
- The dev/test fallback that loads the worker's raw `.ts` source (used
  whenever no compiled `dist/` sibling exists yet - `npm run dev`, or
  vitest running directly against `src/`) now explicitly passes
  `--experimental-strip-types`, since Node's built-in TypeScript stripping
  only became enabled by default in 22.18.0, below this feature's own
  22.13.0 floor.

## Verification

Every fix above was verified by reverting the change (keeping its new
test) and confirming the test genuinely failed against the prior code,
then restoring the fix - not just "test passes now." Full suite is green
(lint, format, build, and the test suite) on Node 18/20/22.

Also verified against a real attachment during development (not included
here): fetched an actual 73KB, 3-page quote PDF via `list-mail-attachments`,
ran it through officeparser standalone, and confirmed the pricing table
came through as clean markdown - 2,089 tokens versus the 24,467 the same
file costs as base64.

Happy to adjust the flag name, the size/timeout/concurrency limits, or
move the dependency handling if you'd rather it work differently.


msoukhomlinov:
@eirikb , whats your view on this one?

eirikb:
Hi

This is quite the PR, and I've seen similar things before, see #373 and #574.  
#373 did drive file text extraction with `pdf-parse` and `mammoth` and in the end I didn't accept it, and in #574 I landed on where I think the line is: this server wraps Graph, Graph has no text extraction for me to wrap, so that part belongs in a separate MCP running alongside it.

Credit where due, this one avoids the registration layer #373 added, it's just an utility tool on the existing path.

Thinner ice for me is mail attachments over HTTP, since a separate MCP can't fetch those itself and `download-bytes-to-file` is stdio only. However `officeparser` is a steep price for it - drags in pdfjs-dist and tesseract, ~130MB, and optional deps install by default so everyone gets that on `npx` even with the flag off. #373 was much lighter than this (in size) and I still said no.

So you see I'm conflicted. On one hand I've said before special handling is out-of-scope, but then we also have special MCP tools now for fetching files simply because MCP + Graph isn't always perfect fit..
I guess these things you introduce here can't be reproduced by simply calling graph?  At least not for non-work users?  
Not that I know if I would combine tools either, as that has also been rejected before.

----

Also, claude said;

Checked what Graph can actually do here:

* `?format=` only offers jpg, pdf and html, and html is limited to loop/fluid/whiteboard files. No text output for Office docs. ([docs](https://learn.microsoft.com/en-us/graph/api/driveitem-get-content-format))
* The Excel workbook API is business only: "Support for workbooks stored in OneDrive Consumer platform is still not available." ([docs](https://learn.microsoft.com/en-us/graph/api/resources/excel))
* `copilot-retrieve` covers drive PDFs and Word for work users, but needs Copilot licensing and is work only.
* Mail attachments have no driveItem behind them, so none of the above reach one.

So personal accounts have no native path for any of these formats, and for work users the gap is attachments plus anything outside Copilot licensing.

Worth noting though that `download-bytes`, `download-bytes-to-file` and `get-download-url` are all transport, they move bytes without interpreting them. `convert-document` would be the first tool here that parses a file format.

msoukhomlinov:
Yeah, it's large - measured it at 161MB installed, once you count `@napi-rs/canvas` (native prebuilt binaries pulled in transitively by `pdfjs-dist`) on top of `tesseract.js-core`.

What do you think of this: I drop officeparser for a per-format bundle instead - `unpdf` (PDF, 2.1MB, zero deps, pure JS) plus a maintained RTF parser for attachments, well under 40MB total. I'd also keep it out of `package.json` entirely and gate it behind a runtime `try/catch` that tells the user to install it themselves if they want the tool, so default installs carry zero extra weight regardless of flag state.

And separately - I'd drop drive-file conversion from this PR. I think your #373/#574 call was right for that half: drive files already have `download-bytes`/`get-download-url`, so text extraction on top of those is buildable outside this server by anyone who wants it. Attachments are the actual gap - no driveItem, no shareable URL, only an OAuth-gated Graph call scoped to this server's session, so nothing outside this server can reach that content at all. That would make this PR attachment-conversion only - smaller diff, smaller dependency, and it's the part of the case you weren't pushing back on anyway.

eirikb:
Hi

Ok I've been thinking, the narrowing sounds right to me. Attachments only, and keeping it out of `package.json`, that addresses most of what I was worried about.

I'm tempted to accept this, before we go there though, one thing I'd like to check. Back in #168 I floated mapping up a special endpoint on this server for a fake direct file download, exactly because attachments are the one thing we can't hand out a download URL for. I never got around to it.

Would that solve your case instead? A short-lived attachment URL served by this server in HTTP mode, so the bytes become reachable from the outside and the extraction can live in whatever MCP you point at it. Then we don't parse anything here, we just do for attachments what `get-download-url` does for drive files.

Not free either of course, it's a new endpoint with expiry and auth to think about. But it keeps this server to moving bytes around, and it would work for any format instead of only the ones we pick a parser for.

What do you think?

msoukhomlinov:
Agreed on the endpoint approach.

Before committing to it I wanted to confirm something could actually consume a download URL. From what I've found so far, the available converter MCPs either take a local filepath for binary formats, or accept a bare URL with no way to send an Authorization header — so I haven't found one that can fetch an authenticated document today. Possible I've missed something.

I'm building that consumer as a separate MIT-licensed MCP server, so the endpoint would have a real caller pointing at it.

On that basis I'll rework #603 to your design: drop `convert-document`, implement the short-TTL attachment URL. Three things I'd rather you decide before I write it:

1. **Where the ticket goes.** Most converters can only fetch a bare URL, so it probably needs to be accepted in the path. I'd also accept it as a bearer for callers that can send headers, though happy to keep it to the path form if you'd prefer fewer moving parts.
2. **TTL and reuse.** My default would be ~120s, single successful redemption, memory-only, and refusing to mint unless `MS365_MCP_PUBLIC_URL` is set — otherwise we hand out URLs pointing at a hostname nobody can resolve.
3. **New PR or amend #603?** Amending keeps the discussion in one place, though the diff would be near-total.

Worth flagging early: streaming at redemption needs a Graph token at redemption. In stdio/MSAL mode we can store the account key and resolve the token on the way through. In OAuth/OBO mode the incoming token is per-request, so the ticket would have to hold the exchanged token for its TTL — memory-only, wiped on redemption or expiry, but it is new material at rest, with the short TTL as the mitigation. Let me know if that changes your view.

Holding off on pushing anything until you've weighed in.

msoukhomlinov:
Superseded by #623, closing.

The approach changed rather than the goal. This PR parsed documents inside the server that holds the mailbox credentials; #623 has the server mint a short-TTL, single-use, HMAC-signed URL instead, and parsing moves to a separate process that never sees a Graph token.

That turned out to be both smaller and better scoped — ~1,400 lines against this PR's ~3,500, because `get-download-url` already existed and only needed its four refusal points extended, rather than a new tool plus a conversion worker.

Thanks for the review time spent here; the concerns raised about buffering and worker preemption were what pushed the design this way.

PR #605, opened 2026-07-27
fix(security): redact PII/tokens in logs by default

msoukhomlinov:
## Summary

`MS365_MCP_REDACT_PII` defaulted to off. That means every Graph API request logged its full options object — including the live OAuth bearer token — to the log file in plaintext, unless an operator explicitly opted in. The redaction logic (JWTs, `Authorization: Bearer` headers, OAuth token fields, auth codes, email addresses) already existed and is wired into the winston logger; it just never ran unless `MS365_MCP_REDACT_PII=true|1` was set.

This server necessarily handles live Microsoft Graph bearer tokens with mail/calendar/contacts read-write scopes, so logging should be secure-by-default with an explicit opt-out, not an opt-in.

## What changed

- `redactionEnabled()` in `src/lib/log-redactor.ts` now returns `true` unless `MS365_MCP_REDACT_PII` is explicitly set to `false` or `0`.
- Updated the doc comments in `src/lib/log-redactor.ts` and `src/logger.ts` to describe the new default.
- Updated `README.md`'s env var table entry for `MS365_MCP_REDACT_PII` to reflect the new default and the opt-out.
- Updated `test/log-redactor.test.ts` to assert the new default (on when unset, off for `false`/`0`, on for anything else).

## Opting out

Operators who want fully verbose logs for local debugging can set `MS365_MCP_REDACT_PII=false` (or `0`).

## Test plan

- [x] `npm run generate && npm run lint && npm run format:check && npm run build && npm test` all pass (660/660 tests, 0 lint errors)
- [x] `test/log-redactor.test.ts` covers unset (redact on), `true`/`1` (on), `false`/`0` (off), and any other value (on)

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,7 +596,7 @@ Environment variables:
 - `MS365_MCP_CLOUD_TYPE=global|china`: Microsoft cloud environment (alternative to --cloud flag)
 - `LOG_LEVEL`: Set logging level (default: 'info')
 - `SILENT=true|1`: Disable console output
-- `MS365_MCP_REDACT_PII=true|1`: Scrub JWTs, Bearer headers, OAuth token fields, and email addresses from log messages before they are written (default: disabled). Useful when logs are shipped to a central store or shared host.
+- `MS365_MCP_REDACT_PII=false|0`: Disable scrubbing of JWTs, Bearer headers, OAuth token fields, and email addresses from log messages (default: enabled). The server handles live Graph bearer tokens, so redaction is on unless you opt out for fully verbose local debugging.
 - `MS365_MCP_CLIENT_ID`: Custom Azure app client ID (defaults to built-in app)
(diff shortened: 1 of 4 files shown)


eirikb:
Hi and thanks.

Good timing, unless it was based on the activity - #604 landed earlier today and does the surgical version of this, strips `accessToken` at the graph-client call site. This one catches everything that fix can't reach (MSAL error strings, `?code=`, Bearer headers), so I think they go well together.

One thing claude highlight: the Bearer pattern matches any word after `Bearer`, not just token-shaped values. So `Bearer token is expired` turns into `Bearer [REDACTED] is expired`, and Graph's `WWW-Authenticate: Bearer realm=""...` gets a bit mangled. Not something you introduced, but it matters more now that it runs by default. Not that I care much, this is nitpick in logs...

Merging this. Thanks

PR #607, opened 2026-07-28
fix(sharepoint): use documented colon route for get-sharepoint-site-by-path

### LINKED ISSUE #606: get-sharepoint-site-by-path always fails: 400 "Error in query syntax" (wrong route form + unencoded-path regression)
`get-sharepoint-site-by-path` returns `400 Bad Request - Error in query syntax` for every parameter shape, so the tool can never succeed.

Version: **0.134.4**, HTTP mode, `--org-mode --read-only --preset excel,outlook,files,search,work`.

## Repro

All three calls fail identically (distinct Graph request-ids, so each really hit Graph):

| `siteId` | `path` | Graph request-id |
|---|---|---|
| `contoso.sharepoint.com` | `/sites/factory` | `208c1bd7-1195-48d3-b34d-9879a3fc9567` |
| `root` | `/sites/factory` | `108abd1d-f419-43ac-8fd9-9331856f72b9` |
| `contoso.sharepoint.com` | `sites/factory` | `9e094817-ca78-4b2e-bd91-05acb91e8683` |
| `<hostname>,<spsite-guid>,<spweb-guid>` | `/factory` | `c60a814e-5c3b-4078-a291-1c120407b7af` |

```json
{"error":{"code":"BadRequest","message":"Bad Request - Error in query syntax."}}
```

The same server, same token, same `Sites.Read.All` scope resolves the site fine via `get-sharepoint-site`, so this is not permissions or auth.

## Cause 1 — the route form isn't the documented v1.0 one

The generated entry is:

```js
{
  method: "get",
  path: `/sites/:siteId/getByPath(path=':path')`,
  alias: "get-sharepoint-site-by-path",
  description: `Invoke function getByPath`,
  response: z.void()
}
```

`getByPath` exists in the OData metadata (hence the auto-generated `Get-MgSiteByPath` cmdlet), but [Get a site resource by path](https://learn.microsoft.com/graph/api/site-getbypath?view=graph-rest-1.0) documents the v1.0 HTTP request as the colon form:

```http
GET /sites/{hostname}:/{relative-path}
```

Requesting the function form against `graph.microsoft.com/v1.0` is what produces `Error in query syntax`.

## Cause 2 — the entry has no `parameters`, so both path params get `encodeURIComponent`'d

This is independent of Cause 1 and would still be a bug if the route were right.

`get-sharepoint-site-by-path` is the only one of its neighbours generated **without** a `parameters` array:

| alias | has `parameters` |
|---|---|
| `get-sharepoint-site` | yes |
| `list-sharepoint-site-drives` | yes |
| `get-sharepoint-site-by-path` | **no** |

With `parameterDefinitions = tool.parameters || []` empty, `paramDef` is `undefined` for both `siteId` and `path`, so execution falls through to the untyped fallback in `executeGraphTool`:

```js
} else if (path2.includes(`:${paramName}`) || ...) {
  const encodedValue = encodeURIComponent(paramValue).replace(/%3D/g, "=");
```

so `path: "/sites/factory"` becomes `%2Fsites%2Ffactory` **inside the quoted OData string literal**:

```
/v1.0/sites/contoso.sharepoint.com/getByPath(path='%2Fsites%2Ffactory')
```

There is already a `skipEncoding` mechanism in `endpoints.json` for exactly this (used for `address`, `q`), but this endpoint doesn't set it.

## Impact

Low severity — `get-sharepoint-site` with a composite id (`contoso.sharepoint.com:/sites/factory`) covers the use case and works today. But the broken tool is the one whose *name* makes it the obvious entry point for path-based lookup, so an agent reaches for it first and burns turns on a call that cannot succeed. It's enabled by default in the `work` preset.

## Suggested fix

Maintainer's call, but two options:

1. **Drop it.** Path addressing is already reachable through `get-sharepoint-site` with a composite id; the tool adds no capability.
2. **Re-point it at the documented colon syntax** — `/sites/:siteId:/:path` — and add `"skipEncoding": ["path"]` so the relative path isn't percent-encoded.

If it stays, worth mentioning the composite-id form in the description, since that's the working path today.

## Workaround for anyone hitting this

```
get-sharepoint-site   siteId = "contoso.sharepoint.com:/sites/<site>"
  → list-sharepoint-site-drives   siteId = <returned id>
  → list-folder-files             driveId = <library driveId>, driveItemId = "root"
```

Verified end-to-end on 0.134.4 (site → library `driveId` → folder listing → `get-excel-range` → `get-download-url`).

stoynovski-a11y:
Fixes #606.

`get-sharepoint-site-by-path` returns `400 Bad Request - Error in query syntax` for every parameter shape, so the tool can never succeed. Two independent defects, both fixed by one `endpoints.json` change.

## 1. The route form

The endpoint was generated from the OData function in the CSDL metadata:

```
/sites/{site-id}/getByPath(path='{path}')
```

[Get a site resource by path](https://learn.microsoft.com/graph/api/site-getbypath?view=graph-rest-1.0) documents the v1.0 request as colon addressing:

```http
GET /sites/{hostname}:/{relative-path}
```

`pathPattern` now points at that form.

## 2. Missing `parameters` → path separators percent-encoded

This one is independent and would still bite on a working route.

`get-sharepoint-site-by-path` was the only one of its neighbours generated **without** a `parameters` array:

| alias | had `parameters` |
|---|---|
| `get-sharepoint-site` | yes |
| `list-sharepoint-site-drives` | yes |
| `get-sharepoint-site-by-path` | **no** |

With `parameterDefinitions = tool.parameters || []` empty, `paramDef` is `undefined` and both params fall through to the untyped fallback in `executeGraphTool`, which `encodeURIComponent`s them — so the relative path became `%2Fsites%2Fmarketing` *inside the OData string literal*.

Because the colon path is absent from Microsoft's published spec, the generator's existing synthesis branch in `simplified-openapi.mjs` now supplies the operation, and derives both path params from the `{...}` placeholders. Regenerating produces:

```js
{
  method: 'get',
  path: '/sites/:siteId:/:path',
  alias: 'get-sharepoint-site-by-path',
  parameters: [
    { name: 'siteId', type: 'Path', schema: z.string().describe('Path parameter: site-id') },
    { name: 'path',   type: 'Path', schema: z.string().describe('Path parameter: path') },
  ],
}
```

Typed `Path` params alone still encode, so the change adds `"skipEncoding": ["path"]` — the same mechanism already used for `address` and `q`.

## Verification

- `npm run generate` → the entry regenerates with both typed path params (shown above).
- `npm test` → **664 passed** (660 existing + 4 new). `npm run lint` clean (warnings pre-existing), `npm run format:check` clean.
- New `test/sharepoint-site-by-path.test.ts` pins both halves: that `pathPattern` is the colon form and never contains `getByPath`, and that the built URL is `/sites/contoso.sharepoint.com:/sites/marketing` with no `%2F`.
- **Negative control** — dropping `skipEncoding` makes the new tests fail with exactly the old broken URL, so the test isn't vacuous:
  ```
  Received: "/sites/contoso.sharepoint.com:/sites%2Fmarketing"
  ```

What I could **not** verify: a live Graph call through a patched build, which needs a full OAuth flow against a modified deployment. What I did confirm against live Graph (v1.0, `Sites.Read.All`, org mode) is that colon addressing resolves the site while the `getByPath` form fails — `get-sharepoint-site` with a composite `hostname:/sites/<name>` id returns 200 for the same site that yields `Error in query syntax` through this tool, across four parameter shapes with four distinct Graph request-ids.

## Known limitation

`path` is expected without a leading slash (`sites/marketing`, not `/sites/marketing`); a leading slash produces `:://` and would fail. The added `llmTip` states the expected shape. Normalising it would mean touching `executeGraphTool` rather than config, so I left it out of this PR — happy to add it if you'd prefer the tool be lenient.

If you'd rather drop the tool entirely — `get-sharepoint-site` with a composite id already covers path lookup — say so and I'll swap this for a removal.

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -1944,11 +1944,13 @@
     "llmTip": "Deletes a column from a SharePoint list. This is irreversible — all data stored in this column across every list item is lost. Confirm with the user before calling. Cannot delete built-in columns (Title, Created, Modified, etc.)."
   },
   {
-    "pathPattern": "/sites/{site-id}/getByPath(path='{path}')",
+    "pathPattern": "/sites/{site-id}:/{path}",
     "method": "get",
     "toolName": "get-sharepoint-site-by-path",
     "presets": ["work"],
-    "workScopes": ["Sites.Read.All"]
+    "workScopes": ["Sites.Read.All"],
+    "skipEncoding": ["path"],
+    "llmTip": "Resolve a SharePoint site from its server-relative URL. site-id is the hostname (contoso.sharepoint.com) and path is the site path without a leading slash (sites/marketing, teams/hr). Returns the site object whose id feeds list-sharepoint-site-drives."
   },
   {
     "pathPattern": "/sites/delta()",
--- a/test/sharepoint-site-by-path.test.ts
+++ b/test/sharepoint-site-by-path.test.ts
@@ -1,0 +1,113 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { z } from 'zod';
+import { readFileSync } from 'fs';
+import { fileURLToPath } from 'url';
+import path from 'path';
+import { registerGraphTools } from '../src/graph-tools.js';
+import type { GraphClient } from '../src/graph-client.js';
+
+vi.mock('../src/logger.js', () => ({
+  default: {
+    info: vi.fn(),
+    error: vi.fn(),
+    warn: vi.fn(),
+  },
+}));
+
+vi.mock('../src/generated/client-beta.js', () => ({ api: { endpoints: [] } }));
+
+// Mirrors what `npm run generate` emits for this endpoint: the synthesized
+// operation carries both path params, because the colon-addressing path is not
+// in Microsoft's published OpenAPI metadata.
+vi.mock('../src/generated/client.js', () => ({
+  api: {
+    endpoints: [
+      {
+        alias: 'get-sharepoint-site-by-path',
+        method: 'get',
+        path: '/sites/:siteId:/:path',
+        description: 'Resolve a SharePoint site from its server-relative URL.',
+        parameters: [
+          { name: 'siteId', type: 'Path', schema: z.string() },
+          { name: 'path', type: 'Path', schema: z.string() },
+        ],
+      },
+    ],
+  },
+}));
+
+const endpointsPath = path.join(
+  path.dirname(fileURLToPath(import.meta.url)),
+  '..',
+  'src',
+  'endpoints.json'
+);
+const endpoints = JSON.parse(readFileSync(endpointsPath, 'utf8')) as Array<{
+  toolName: string;
+  pathPattern: string;
+  skipEncoding?: string[];
+}>;
+
+describe('get-sharepoint-site-by-path', () => {
+  const entry = endpoints.find((e) => e.toolName === 'get-sharepoint-site-by-path')!;
+
+  it('uses the documented v1.0 colon-addressing route, not the getByPath function', () => {
+    // GET /sites/{hostname}:/{relative-path} per
+    // https://learn.microsoft.com/graph/api/site-getbypath
+    // The OData function form `/sites/{site-id}/getByPath(path='{path}')` is in the
+    // CSDL metadata but returns 400 "Error in query syntax" against v1.0.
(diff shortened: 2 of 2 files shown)


eirikb:
Merging this. Thanks 🙂

PR #608, opened 2026-07-28
feat: support Sites.Selected for direct SharePoint tools

sharkeyl:
## What changed

This adds support for using `Sites.Selected` with the direct SharePoint tools. Softeria's `--allowed-scopes` feature exists precisely for enterprise deployments, and `Sites.Selected` is a standard Microsoft least-privilege pattern. This is a gap in the current codebase.

The main use case is tenants that do not want to grant broad SharePoint permissions like `Sites.Read.All`, but still want the MCP server to work against specific SharePoint sites that have been approved for the app.

Existing broad-scope setups should keep working as they do today. I’ve left those as the default login path.

## A couple of details

This only applies to direct SharePoint site/list/item/column tools, where the caller already knows which site they want to work with.

Search and tenant-wide discovery tools still require the broader SharePoint scopes, because `Sites.Selected` does not give the app a tenant-wide site search surface.

## Checked

- `npm run generate`
- `npm test -- --run test/or-group-scopes.test.ts test/scope-derivation.test.ts test/endpoints-validation.test.ts`
- `npm run format:check`
- `npm run build`
- `npm run lint`
- `npm test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,21 @@ CLI value takes precedence over `MS365_MCP_ALLOWED_SCOPES`; if neither is set, t
 
 Scope coverage is hierarchy-aware: for example, `Mail.ReadWrite` covers tools that require `Mail.Read`, and `Files.ReadWrite.All` covers tools that require `Files.Read`.
 
+SharePoint supports two enterprise permission models:
+
+- Broad tenant scopes such as `Sites.Read.All`, `Sites.ReadWrite.All`, and `Sites.Manage.All`.
+- Microsoft Graph `Sites.Selected`, where SharePoint site access is granted to the app on specific site collections and Graph evaluates the signed-in user's own permissions at request time.
+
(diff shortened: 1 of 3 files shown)


eirikb:
Hi and thanks :slightly_smiling_face: 

Nice use of the OR-group scopes from #536, exactly what they're for.

Tried to verify it myself but couldn't - my old consent grant means the token still carries `Sites.Read.All` alongside `Sites.Selected`, so everything passes regardless. Did you run it against a site the app was actually granted on?

If you fix the conflict I will merge.

sharkeyl:
Hi @eirikb, thank you for taking a look!

I'll resolve the conflict and test again to confirm all is well, and get back to you later today with evidence of testing :)

sharkeyl:
I ran a couple of checks.

First I tested the scope filtering locally with `--list-permissions`:

  - `Sites.Selected` -> direct SharePoint tools are enabled
  - `Sites.Read.All` -> still works as before
  - `User.Read` only -> direct SharePoint tools are disabled

So the OR-group handling looks right.

I also ran it through the MCP server against Graph, using direct site/drive calls rather than search. These all succeeded:

  - `get-sharepoint-site-by-path`
  - `list-sharepoint-site-drives`
  - `get-drive-root-item`
  - `list-sharepoint-site-lists`

The client also reported that no tenant-wide SharePoint search/discovery tool was available.

I don't manage the SharePoint site grants myself, so I can't independently confirm the site permission object, but the token used for this test does not have `Sites.Read.All` / `Sites.ReadWrite.All` / `Sites.Manage.All`.

PR #609, opened 2026-07-29
feat(audit): include target resource metadata for direct Graph calls

sharkeyl:
## Summary

Adds `target_resource` to audit log entries for generated Graph tools where the server can work out a specific resource from the Graph path.

This makes the audit log a bit more useful for hosted deployments: it already records who called a tool and whether it worked; this adds which Graph resource was involved for direct resource calls.

A couple of things this deliberately avoids:

- logging tool params
- logging response bodies or file contents
- adding fake targets for broad list/search calls

For `/content` and `/$value` endpoints, the audit entry points at the parent resource.

I also added coverage for success and error cases.

## Testing

```sh
npm test -- --run test/audit-target-resource.test.ts src/__tests__/graph-tools.test.ts
npm run format:check
npm run build
npm run lint
```


sharkeyl:
Redacted logs below. Broad/setup calls don’t include `target_resource`, as expected:

```json
{"event":"tool.call","http_method":"GET","request_id":"3022b2a1-6aad-4e45-ae83-7848dcafabbe","status":"success","tool":"list-drives","user_principal_name":"<redacted>","duration_ms":443}
{"event":"tool.call","http_method":"GET","request_id":"b302f644-f3c5-451c-bb6e-2e50ec00571a","status":"success","tool":"get-drive-root-item","user_principal_name":"<redacted>","duration_ms":385}
```

Specific resource access does include it:

```json
{"event":"tool.call","http_method":"GET","request_id":"18cc62d8-0a1b-4333-a68f-8674f56451e9","status":"success","tool":"get-drive-item","user_principal_name":"<redacted>","duration_ms":381,"target_resource":{"type":"drive_item","id":"/drives/<drive-id>/items/<item-id>"}}
```

And it’s still present when Graph returns an error:

```json
{"event":"tool.call","http_method":"GET","request_id":"4cc5769c-a7a7-41a7-a30d-167734b1f3ac","status":"error","tool":"get-drive-item","user_principal_name":"<redacted>","duration_ms":380,"target_resource":{"type":"drive_item","id":"/drives/<drive-id>/items/audit-target-resource-test-missing-item"}}
```

So the behaviour looks right to me: broad calls stay as they were, and specific Graph resource calls now get target_resource on both success and failure. I didn’t test the SharePoint-specific target types here. In our setup, users can’t browse/search across SharePoint to find arbitrary sites; access is limited to approved site grants. So I don’t have a clean way to pick a site from the client just to exercise those paths. The drive-item calls above are real Graph calls and cover the important behaviour: `target_resource` is added for specific resource access on both success and failure.

eirikb:
Hi, and thanks for this!

Two things before merging.

The drive path block puts filenames into the audit log, and that log is on by default and skips the redaction from #605. Could you drop it? The id-addressed ones cover the same files anyway.

And the regex table, could that just be the last `{...-id}` in the pathPattern? So `/me/messages/{message-id}` gives `message`. Would be less code, and nothing to touch when I add endpoints later. Might be I'm missing a case though.

What do you think?

---

Also, claude said;

* Coverage today is 46 of the 182 pathPatterns that address a specific resource. `/me/messages/{message-id}`, all of `/groups/{group-id}/*` and `/chats/{chat-id}/messages/{chatMessage-id}` get nothing, while the attachment and hostedContents children of those same paths do get a target. The 57 patterns with no id param are correctly skipped.
* The `pathPattern` / `params` branch in `deriveTargetResource` never runs. `resolvedPath` is always passed from `graph-tools.ts` and always starts with `/` (checked all 309 generated paths), so the `??` short-circuits every time. That's `resolveGraphPathForAudit`, `valueForPlaceholder`, `toCamelCase`, `toKebabCase` and `CONTROL_PARAM_NAMES`, around 55 lines.
* Side effect of that: the unresolved-placeholder guard only exists on the dead branch, so `/drives/d1/items/:driveItemId` gets logged as a drive_item id if a path param ever fails to substitute.
* On the last-param idea, the `-id` suffix also drops `{q}`, `{address}` and the `root:/{path}:` form, so the filename thing above goes away on its own.
* The reason the string parsing is awkward at all is on my side, not yours: `executeGraphTool` has the pattern, the param name and the value at the point it substitutes them, then flattens it all into a string and appends the query before the audit call. So the info exists, it just gets thrown away a few lines too early.

`stdout` -> `stderr` in the docs is a good catch, that was wrong.

sharkeyl:
Hi @eirikb, thanks for the review!

Runtime smoke test:

```json
{"event":"tool.call","http_method":"GET","request_id":"<redacted>","status":"success","tool":"list-drives","user_principal_name":"<redacted>","duration_ms":286}
{"event":"tool.call","http_method":"GET","request_id":"<redacted>","status":"success","tool":"get-drive-root-item","user_principal_name":"<redacted>","duration_ms":395,"target_resource":{"type":"drive","id":"/drives/<drive-id>"}}
{"event":"tool.call","http_method":"GET","request_id":"<redacted>","status":"success","tool":"get-drive-item","user_principal_name":"<redacted>","duration_ms":395,"target_resource":{"type":"drive_item","id":"/drives/<drive-id>/items/<item-id>"}}
{"event":"tool.call","http_method":"GET","request_id":"<redacted>","status":"error","tool":"get-drive-item","user_principal_name":"<redacted>","duration_ms":377,"target_resource":{"type":"drive_item","id":"/drives/<drive-id>/items/audit-target-resource-test-missing-item"}}
```

`list-drives` stays unchanged, `/drives/{drive-id}/root` now logs the drive, and `/drives/{drive-id}/items/{driveItem-id}` logs the drive item on both success and failure.

I didn’t try to force the other cases through the live tenant. In our setup, users can’t just browse/search SharePoint to find arbitrary sites, access is limited to approved site grants, and I avoided enumerating mail or document content purely to find test IDs.

Those cases are covered in tests instead:

- path-style drive access like `/me/drive/root:/{path}:/content` is skipped, so filenames aren’t logged
- `/me/messages/{message-id}` gives target_resource.type="message"
- `get-sharepoint-site-by-path` logs the site identifier, not the `{path} `value
- missing ID params produce no `target_resource`
- non-ID params like `{path}`, `{q}`, and `{address}` are ignored

Validation:

```bash
npm test -- --run test/audit-target-resource.test.ts src/__tests__/graph-tools.test.ts
npm run format:check
npm run build
npm run lint
```

I will be on annual leave until Tuesday 4th August. If you have other feedback, I'll address these on my return 🙂

eirikb:
That's exactly it, thanks!

Coverage went from 46 to 179 of the pathPatterns, and the filename thing fell out by itself like we hoped. Merging this.

Only loss is `/me/photo`, no id in the path so it gets nothing now. Doesn't matter much, the tool name says it anyway. I'll add it myself if I end up missing it.

Enjoy the leave :slightly_smiling_face:

sharkeyl:
Amazing, thank you for your quick responses and feedback on both PRs @eirikb! I hope you have a good weekend ahead 🙂

PR #618, opened 2026-08-05
fix(calendar): clarify time format, UTC/DST guidance, and calendarId/eventId distinction

msoukhomlinov:
## Summary

Cleans up several calendar tool descriptions in `src/endpoints.json` that were misleading or under-specified for LLM callers, plus a few unrelated inaccuracies found while spot-checking other tool descriptions for the same class of issue.

**Calendar tools** (`create-calendar-event`, `update-calendar-event`, `create-specific-calendar-event`, `update-specific-calendar-event`, `delete-specific-calendar-event`):
- Clarify that event times use nested `{dateTime, timeZone}` objects, not flat `startDateTime`/`startTimeZone` fields (the latter are silently dropped, not rejected).
- Correct the UTC guidance: UTC is simplest for one-off events, but Graph resolves recurrence/DST against whatever zone is set on `start.timeZone` — forcing UTC breaks DST handling for recurring events. Points models at the existing `get-mailbox-settings`/`list-supported-time-zones` tools to get a valid zone name instead of guessing one from memory.
- Removed a sentence duplicated between `descriptionOverride` and `llmTip` on `delete-specific-calendar-event`.
- Made `calendarId` wording consistent across the four calendar tools that take it.

**Other tools**, found via a spot-check across mail, files/OneDrive, Teams, Planner/ToDo/OneNote, and Excel/contacts/users descriptions against the Graph OpenAPI schema:
- `extract-drive-item-sensitivity-labels`: response has no `value` wrapper — `labels` is a root property.
- `update-planner-task`: priority preset mapping was shifted a full band (correct Planner UI presets are 1=Urgent, 3=Important, 5=Medium, 9=Low).
- `list-relevant-people`: person schema field is `scoredEmailAddresses`, not `emailAddresses`.
- `get-mail-message-mime`: description and llmTip cited different RFC numbers (822 vs 5322) for the same MIME content; standardized on 5322.
- `create-online-meeting`: only `endDateTime` is documented as required, not `subject`/`startDateTime`.

## Test plan
- [x] `npx vitest run test/discovery-search.test.ts test/endpoints-validation.test.ts` — 51/51 pass
- [x] `npm run lint` — 0 errors (pre-existing warnings only, unrelated to this diff)
- [x] `python3 -c "import json; json.load(open('src/endpoints.json'))"` — valid JSON

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -364,7 +364,7 @@
     "toolName": "create-calendar-event",
     "presets": ["calendar", "outlook", "personal"],
     "scopes": ["Calendars.ReadWrite"],
-    "descriptionOverride": "Create (schedule) a new calendar event — a meeting or appointment — on the user's calendar. Times use nested objects, not flat fields: start: {dateTime, timeZone}, end: {dateTime, timeZone}. Pass times in UTC (e.g. 3:30 PM AEDT = 04:30 UTC). Do NOT use startDateTime/startTimeZone — those flat fields are ignored. Set subject, location, body, and attendees; supports online meetings and recurrence.",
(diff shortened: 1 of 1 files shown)


eirikb:
Nice cleanup, thanks :slightly_smiling_face: 

One thing before I merge: the `create-online-meeting` tip. I think `subject` is required too, the [onlineMeeting resource docs](https://learn.microsoft.com/en-us/graph/api/resources/onlinemeeting?view=graph-rest-1.0) say "Required when you create an online meeting" on both `subject` and `endDateTime`. Could you revert that one? Or am I wrong?

Rest looks good to me.

eirikb:
Merged, thanks :slightly_smiling_face: 

I pushed a commit on top before merging. `subject` does carry the same "Required when you create an online meeting" note as `endDateTime`, so I put that one back. And `extractSensitivityLabels` went the other way too - the action page response example does have the `value` wrapper, even if the resource type page shows `labels` at the root. A bit confusing, but I think the action page is the one that counts for us.

Rest was spot on, the planner priority mapping especially

PR #619, opened 2026-08-06
fix: Disable SSE, since MCP is stateless

joelw:
This aligns the MCP with the stateless HTTP example in the modelcontextprotocol SDK: https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.29.0/src/examples/server/simpleStatelessStreamableHttp.ts

Instead of opening up an SSE channel, reply to `GET /mcp` with 405 Method not allowed. This resolves an [issue in LibreChat](https://github.com/danny-avila/LibreChat/discussions/9621) which goes into an endless loop of restarting SSE which causes lots of useless log messages.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -701,50 +701,16 @@ class MicrosoftGraphServer {
         allowUnauthenticatedDiscovery: this.options.allowUnauthenticatedDiscovery,
         publicUrl: publicBase,
       });
-      app.get(
-        '/mcp',
-        mcpAuth,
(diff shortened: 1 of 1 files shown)


eirikb:
Hi and thanks. Looks good to me. Should not break the HTTP version of the server and remove the thing we didn't support anyway.

Note: GitHub had an outage and the build for this PR got messed up - I'm going to try to re-trigger it by closing and re-opening.

eirikb:
Ha - it worked!

joelw:
Awesome! Thanks @eirikb for the merge and for your tireless work on this MCP.

PR #623, opened 2026-08-07
feat: mint short-TTL signed download URLs for Graph bytes that have none

msoukhomlinov:
Supersedes #603. That PR added `convert-document`, which parsed attachments **inside** the server that holds the mailbox credentials. This takes the opposite approach: the server never parses anything, it just hands out a short-lived URL that something else can fetch.

## What it does

`get-download-url` already returns Graph's pre-authenticated `@microsoft.graph.downloadUrl` for drive and SharePoint items. It refuses for everything else, because mail attachments, meeting recordings and other `$value` endpoints have no such URL — they are only reachable with an `Authorization` header.

This closes that gap by minting one: a single-use, short-TTL, HMAC-signed URL served by this server, which streams the Graph bytes when redeemed.

- **Off by default.** Requires `--enable-attachment-urls`, HTTP mode only. stdio warns and ignores.
- **Grants no authority the caller did not already have.** Minting is refused whenever Graph identity comes from the request (bearer, OBO, and `MS365_MCP_OAUTH_TOKEN` OAuth mode) — otherwise a ticket minted as one identity could be redeemed as whatever account the server had cached.
- **Single-use, short-TTL.** Default 120s, hard maximum 300s. Redeeming burns the ticket. The store refuses at capacity rather than evicting, so a mint loop cannot invalidate someone else's live ticket.
- Uniform 404 for unknown, spent and expired tickets — no oracle distinguishing them.

## Two ports are not two surfaces unless they bind two interfaces

`--attachment-port` puts the attachment route on its own listener; `--attachment-host` binds it to its own interface.

The second flag is the one that matters, and it is easy to omit. In a container deployment, network membership grants access to **every port on a container, not one port**. A sidecar placed on a shared bridge purely to fetch attachments can equally dial `/mcp` — and with `--trust-proxy-auth`, where reachability *is* the authentication, that is the entire tool surface with no credential. Splitting the port isolates routes; only splitting the interface isolates callers.

The dedicated listener is deliberately bare: no body parsers, no CORS, no OAuth router, no `/mcp`, not even the health check. Everything else 404s by construction rather than by a rule that has to keep matching. `trust proxy` is left off there, so the one uncredentialed surface cannot pick its own rate-limit bucket via `X-Forwarded-For`.

## Incidental fix

Express 5 registers the `app.listen` callback as the server's `error` handler, so a failed bind *invoked* the success callback: the server logged that it was listening on a port it had not got, and stayed up serving nothing. `listen()` is now awaited and rejects properly.

## Tests

732 passing across 59 files (+107 over the base), including the signing wire format pinned against verified vectors, the identity-escalation regression, and interface isolation asserted with real TCP on distinct loopback addresses — cross-interface dials assert *connection refused*, and the cross-port case additionally asserts the ticket is unburnt and still redeemable, since a bare 404 would also be produced by a handler that ran and refused.

`npm run generate && lint && format:check && build && test` all clean (0 errors; pre-existing warnings unchanged).

---

The signing format is shared with an external converter, so the canonical string, reserved parameters and base64url digest are specified precisely in `src/lib/url-signing.ts` — including three documented interop traps between Python's `urllib` and the WHATWG `URL`/`encodeURIComponent` behaviour.


eirikb:
Hi there

Thanks for closing #603 and rethinking the whole thing - handing out a URL instead of parsing inside the server is a much better fit here. And the `app.listen` catch was a real bug, nice one.

Two small things before I merge:

1. `downloadStream` reads `content-length` straight off the response, but when Graph doesn't send that header it comes out as `0` instead of unset, so the response gets served with `content-length: 0` and the stream breaks. Needs a null check I think.

2. `get-download-url`'s description still tells the model that mail attachments and recordings have no URL and to use `download-bytes`. That's the exact case you now handle, so with the flag on I bet the model just routes around the new path.

Rest looks good to me.

HenkieTenkie62:
Sounds exactly like something I need. Will this be implemented [msoukhomlinov](https://github.com/msoukhomlinov)?

eirikb:
Hi @msoukhomlinov

Merged main in since it had gone conflicting, and pushed a few commits on top, hope that is ok.

Your two from last time are in. Found a few more while I was there: that same `content-length` was wrong for gzipped bodies too, targets with a `#` or `..` could resolve somewhere else (couple of them reached `/beta`), `HEAD` burned the ticket, and an upstream `inline` came straight through on `content-disposition`. Plain `--http` warns now as well, since the identity guard means it only really mints under `--trust-proxy-auth`.

The path check took me three goes. I kept writing patterns and the URL parser kept finding a spelling I had not thought of.

One I left alone: the signing key is required, but the route ignores the signature, and I think you are right to ignore it. So someone fetching with curl still needs a shared secret for a check nothing on their side runs. Could we make it optional? The IPv6 refusal could probably go with it.

Nice feature though, the ticket model especially. Shout if you disagree with any of it.

PR #633, opened 2026-08-11
feat(teams): offer Chat.ReadBasic as least-privilege scope alternative for list-chats/get-chat

### LINKED ISSUE #630: list-chats/get-chat should offer Chat.ReadBasic as a least-privilege scope alternative (OR-group support already exists, unused)
## Context

I'm building a **write-only Teams deployment**: a custom Azure app whose delegated grants deliberately exclude anything that can read message content, paired with `--enabled-tools` + `--allowed-scopes`. Sending works with just `ChatMessage.Send`, but *finding conversations* is currently impossible in this setup:

- `list-chats` / `get-chat` declare `workScopes: ["Chat.Read"]` in `src/endpoints.json`, so `--allowed-scopes` without `Chat.Read` disables them.
- Microsoft Graph's documented **least-privileged** scope for `GET /me/chats` and `GET /chats/{id}` (including `$expand=members`) is **`Chat.ReadBasic`** — chat topic/type/members, no message bodies. That's exactly the "address book" capability a send-only agent needs.

## The interesting part

The runtime **already supports scope alternatives**: `EndpointConfig.scopes`/`workScopes` are typed `string[] | string[][]`, and `getEndpointEffectiveLoginScopes()` in `src/auth.ts` picks the first OR-group fully covered by the allowlist (with the primary group used by default, per the least-privilege comment there). As far as I can tell **no endpoint in `endpoints.json` actually uses the OR-group form today**, so this feature is dormant.

## Proposal

Change the two entries to:

```json
"workScopes": [["Chat.Read"], ["Chat.ReadBasic"]]
```

- Default users are unaffected (primary group first → still requests `Chat.Read`, same data, e.g. `lastMessagePreview` keeps working).
- `--allowed-scopes '... Chat.ReadBasic ...'` users get `list-chats`/`get-chat` back with a token that cannot read message content. Graph rejects `$expand=lastMessagePreview` etc. under ReadBasic, which is the point.

`list-chat-members` (`ChatMember.Read`) could get the same treatment (`Chat.ReadBasic` is also its least-privileged documented scope), but the two chat-listing tools are the ones that matter for the send-only use case.

nebkat:
Closes #630.

`list-chats` and `get-chat` currently hard-require `Chat.Read`, which makes them unusable in send-only deployments where the Azure app deliberately has no message-content read grants. Microsoft Graph's documented least-privileged scope for both operations (including `$expand=members`) is `Chat.ReadBasic` — chat topic/type/members, no message bodies.

The OR-group machinery for exactly this (`scopes?: string[] | string[][]`, `getEndpointEffectiveLoginScopes` picking the first group covered by `--allowed-scopes`) already exists in `src/auth.ts` but no endpoint used it. This is the first use — a data-only change:

```json
"workScopes": [["Chat.Read"], ["Chat.ReadBasic"]]
```

**Behavior (verified with `--list-permissions` on the built package):**

| Configuration | Before | After |
|---|---|---|
| `--preset teams` (default) | requests `Chat.Read` | unchanged — primary group first |
| `--allowed-scopes '… Chat.Read …'` | `Chat.Read`, tools enabled | unchanged |
| `--allowed-scopes '… Chat.ReadBasic …'` (no `Chat.Read`) | **tools disabled** | tools enabled, requests `Chat.ReadBasic` |

Under `Chat.ReadBasic`, Graph itself rejects message-content access (`$expand=lastMessagePreview`, `/messages`) — which is the point for write-only deployments.

`npm run verify`: 813/813 tests pass, lint/format clean.

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -1560,14 +1560,14 @@
     "method": "get",
     "toolName": "list-chats",
     "presets": ["teams", "work"],
-    "workScopes": ["Chat.Read"]
+    "workScopes": [["Chat.Read"], ["Chat.ReadBasic"]]
   },
   {
     "pathPattern": "/chats/{chat-id}",
     "method": "get",
     "toolName": "get-chat",
     "presets": ["teams", "work"],
-    "workScopes": ["Chat.Read"]
+    "workScopes": [["Chat.Read"], ["Chat.ReadBasic"]]
   },
   {
     "pathPattern": "/chats/{chat-id}/members",


eirikb:

PR #634, opened 2026-08-11
feat: wrap outgoing agent-authored messages in a configurable signoff (opt-in)

### LINKED ISSUE #632: Feature request: configurable message signoff for Teams sends (default 🤖) — mark agent-sent messages
## Context

Every Teams message sent through this server is AI-mediated by construction — an agent composed it, even when a human asked for it. Recipients currently have no way to tell such messages apart from ones the account owner typed themselves.

I currently enforce a marker client-side (a Claude Code PreToolUse hook that rewrites the `send-chat-message` input to append 🤖), but that protection lives in one client's config. The server is the right chokepoint: it sees every outgoing message regardless of which MCP client is attached.

## Proposal

A **message signoff** appended to the body content of every outgoing chat/channel send and reply (`send-chat-message`, `reply-to-chat-message`, `send-channel-message`, `reply-to-channel-message`):

- `--message-signoff <string>` / `MS365_MCP_MESSAGE_SIGNOFF` — the string to append (separated by a space; for `contentType: html` bodies appended after the content, which renders fine).
- **Default: `🤖`** — honest-by-default marking of agent-sent messages; anyone who disagrees flips one flag.
- `--no-message-signoff` / `MS365_MCP_MESSAGE_SIGNOFF=""` — disable entirely.
- Idempotent: skip appending when the content already contains/ends with the configured signoff, so it composes with client-side hooks like mine without doubling.

A default-on behavior change is the debatable part; if that's unacceptable, default-off with the flag documented under a "transparency" section would still be a big improvement — but I'd argue AI-sent-message marking is the right default for this class of tool, for the same reason `--read-only` and the destructive-op confirm gate exist: safe defaults, explicit opt-out.

Happy to send a PR.

nebkat:
Closes #632.

Wraps outgoing agent-authored message content in a configurable signoff, enforced at the Graph request layer (`GraphClient.performRequest`) keyed on method + path rather than tool alias (per review), so every route to a message write passes through the same gate:

- Teams sends and replies (`send-chat-message`, `reply-to-chat-message`, `send-channel-message`, `reply-to-channel-message`) — fail-closed: a send without a signable `body.content` string is refused rather than sent unsigned
- the PATCH tools that could rewrite an already-signed message (`update-chat-message`, `update-channel-message`, `update-channel-message-reply`)
- mail drafts, signed where their content is written (`create-draft-email`, `update-mail-message`, `update-shared-mailbox-message`, and the `comment` / inline message of `createReply`/`createReplyAll`/`createForward`) — `send-draft-message` carries no body at send time, and a user-authored draft sent as-is stays unsigned by design
- direct mail sends, fail-closed like a Teams send: `send-mail` / `send-shared-mailbox-mail` (`message.body.content`), the direct `reply`/`replyAll`/`forward` actions and their shared-mailbox variants (`comment` + inline message; a bare forward passes, the quoted original is not agent-authored), and `reply-to-group-thread` (`post.body.content`)
- `graph-batch` sub-requests, matched recursively against the same rules (query strings and stray version prefixes stripped)

Configuration (opt-in, default off):

- `--message-signoff-prefix <text>` / `MS365_MCP_MESSAGE_SIGNOFF_PREFIX` — e.g. `🤖`
- `--message-signoff-suffix <text>` / `MS365_MCP_MESSAGE_SIGNOFF_SUFFIX`
- `--no-message-signoff` disables both, overriding the env vars

Rationale: a message sent through an agent should not be able to pass as hand-typed by the account owner, so a compromised or prompt-injected agent cannot phish colleagues from behind a trusted name. For a local stdio server this is a guardrail against an agent misusing the tools it was given, not a hard security boundary (the README says so); in http mode it acts as an operator policy control.

Notes:

- Markers already present are not stacked (compared on rendered text for html bodies), so an agent edit round-trip does not accumulate signoffs.
- Html content is normalized before a suffix is appended (`parse5`, promoted to a direct dependency) so an unterminated construct like a trailing `<!--` cannot swallow the marker. A prefix cannot be affected by later content, which is why it is the suggested position.
- Markup in a marker is allowed (e.g. a coloured `<span>`), but a marker that renders as empty text — such as `<agent>` — is rejected at startup rather than silently producing unsigned-looking messages.
- Not yet covered: `forward-calendar-event` and event invite/accept/decline comments (invite semantics) — follow-up.

Tested with unit and Graph-client-level tests (849/849 passing, lint/format clean); the original Teams send path was also verified end-to-end against live Graph.


eirikb:
Hi @nebkat and thanks

Like the idea, but I wonder if we should think about the framing needs before the code.

`graph-batch` takes arbitrary sub-requests, so an agent can POST `/chats/{id}/messages` straight through it and `applyMessageSignoff` never runs, since it keys on `tool.alias`. Same with `update-chat-message` and friends, delegated PATCH can rewrite `body.content`. So more aliases won't fix it.

The bigger thing though: for a local stdio server the agent and the server are the same user on the same machine. It can read the token cache, clear the env var and restart, or just call Graph directly. So we can't really stop a compromised agent there no matter where the check sits. Your PreToolUse hook is about as strong, it's only a per-client hassle.  
Locally, if I wanted something like this, I'd probably add it as a note to a skill instead.

Where it does bind is http mode, since there the operator isn't the agent and it genuinely can't be turned off - IF not same machine. So I think this is an operator policy feature more than a security control, and it'd be good to say so in the README so nobody runs it locally thinking it stops anything.

If we go that way I'd suggest default off, and enforcing in `performRequest` on method + path instead of tool alias. That covers batch, and it scales to the rest - `send-mail`, the reply/forward tools, event invites and accept/decline comments have the same problem, and `send-draft-message` has no body at all so there's nothing to sign at send time.

Also the prefix goes into html unescaped, so a marker like `<agent>` renders as nothing.

What do you think? Would like to land some version of this.

nebkat:
Thanks — I've reworked the PR along the lines you suggested (single commit on top of the original). One thought on the local stdio case first, since I think there's a bit more value there than a skill note would capture:

**On the local threat model.** Agreed that nothing in the server binds an agent with shell access — it can read the cache, restart the server with the env var cleared, or curl Graph itself. But that may be the wrong adversary to design this feature around, for a few reasons:

- The realistic failure mode isn't a compromised process running an evasion playbook — it's a prompt-injected agent following instructions through the tools it was given. A skill note is advisory to exactly the component that's been subverted; a model that's been talked into impersonating its operator can be talked out of an instruction, but it can't be talked out of a deterministic gate on the wire path. Those fail in fundamentally different ways.
- Plenty of MCP clients have no shell at all — Claude Desktop without extra servers, mobile/web harnesses, anything deliberately locked down. For those, stdio or not, the server-side gate genuinely binds: the tools are the only route to Graph the agent has.
- Your suggested fix mostly resolves this anyway: with enforcement in `performRequest`, evasion *through MCP* becomes impossible, and what's left (shell, token theft) is out of scope for any in-server control. "Guardrail against tool misuse, boundary in http mode" is how the README now frames it — no claims beyond that.

So I'd lean towards landing it as a real feature with an honest threat-model note. On to what changed:

**Enforcement moved to `performRequest`, keyed on method + path.**
Tool aliases no longer participate. One refinement to "that covers batch": the outer request is just `POST /$batch`, so the gate parses batch payloads and applies the same rules to each sub-request (recursively, with query strings and stray `/v1.0`/`/beta` prefixes stripped). A matching sub-request that can't carry the signoff fails the whole batch closed.

**PATCH is covered.**
`update-chat-message` / `update-channel-message` / `update-channel-message-reply` re-sign a rewritten `body.content` (a PATCH not touching the body passes through). That made dedup necessary: markers already present aren't re-added — compared on rendered text for html bodies, since Graph returns stored content re-wrapped in markup — so a GET → edit → PATCH round-trip doesn't stack signoffs.

**`send-draft-message`:**
Rather than blocking it or forcing a fetch-and-edit at send time, draft content is signed where it's written — `create-draft-email`, `update-mail-message`, `update-shared-mailbox-message`, and the `comment` / inline message of `createReply`/`createReplyAll`/`createForward` (plus the `mailFolders` / `/users/{id}` create routes that are only reachable through batch). A draft the user wrote themselves and merely asks the server to send goes out untouched — that's arguably not agent authorship.

**Default is now off:**
As you suggested — `🤖` is the documented example, so nothing changes for existing deployments until the env vars or flags are set.

**Escaping:**
I went a different way than escaping the marker. Markup in a marker is deliberately allowed (a coloured `<span>` is a legitimate use), but a marker that renders as empty text — your `<agent>` example — is rejected at startup, since that config would silently produce unsigned-looking messages while the operator believes the feature is on. Fail-open config refused, intentional styling preserved.

Not covered yet: `send-mail`, the direct `reply`/`forward`/`replyAll` tools, and event invite / accept / decline comments. Each is one rule in the table now, so happy to add them here or in a follow-up — I'd lean follow-up to keep this one reviewable.

eirikb:
Good rework, this is much closer 🙂

And fair point on the threat model. Prompt injection through the tools is a different thing from a process with a shell, and the `performRequest` gate closes the part we can actually close. Happy with the README framing now.

Only thing I'd like before merging: `send-mail`. A draft written through `create-draft-email` gets signed, but `send-mail` goes straight out unsigned, and that's the tool people actually use. Same with `send-shared-mailbox-mail` and the direct reply/forward/replyAll. I know you said follow-up, but a half covered signoff feels worse than none to me, since you can't really tell from the outside which path a message took. If they're one line each in the table now, I'd rather take them here.

Rest looks good to me.

nebkat:
Agreed — your point that a recipient can't tell which path a message took is the whole rationale of the feature, so half-coverage undercuts it. All in now:

**`send-mail` / `send-shared-mailbox-mail`:** `POST /(me|users/{id})/sendMail` signs `message.body.content`, fail-closed like a Teams send — a direct outbound send must carry signable content. (Side effect: a body-less, subject-only send is refused while a signoff is configured; an empty-string body is signable, so agents aren't blocked, but nothing goes out unsigned.)

**Direct `reply`/`replyAll`/`forward`** (plus the `/users/{id}` shared-mailbox variants): same applier as the `createReply` draft tools — the `comment` and any inline `message.body` get signed; a bare forward passes, since the quoted original isn't agent-authored.

**One you didn't name but the same logic catches:** `reply-to-group-thread` (`POST /groups/{id}/threads/{id}/reply`) is a direct outbound message too, so it's covered as well (`post.body.content`, fail-closed). `forward-calendar-event` and the event accept/decline comments are the one class still out — invite semantics are different enough that I'd keep them for the follow-up.

All of these ride the same rules table, so `$batch` sub-requests are covered automatically. 849/849 tests.

eirikb:
Awesome! And nice catch on the group thread reply, I missed that one.

Agree on leaving the calendar invite comments for a follow-up. Merging this 🙂

PR #635, opened 2026-08-11
feat(teams): add teams-write preset - send-only counterpart to --read-only

### LINKED ISSUE #631: Feature request: teams-write preset — send-only counterpart to --read-only
## Context

`--read-only` exists for the "let the agent look but not touch" deployment. The mirror-image deployment — **"let the agent send but not read"** — currently requires hand-rolling a long `--enabled-tools` regex.

That write-only shape is a natural fit for org deployments of AI agents on Teams: the agent can notify people and post updates, but a compromised or misbehaving agent (or prompt injection riding on tool output) cannot exfiltrate conversation history, because the enabled tool set (and ideally the Azure app's delegated grants, see #630) never includes message-content read.

## Proposal

A `teams-write` preset (alongside the existing `teams` preset in `endpoints.json` / `tool-categories.ts`) containing roughly:

- `send-chat-message`, `reply-to-chat-message`
- `send-channel-message`, `reply-to-channel-message`
- basic addressing/enumeration only: `list-chats`, `get-chat` (see #630 for making these work under `Chat.ReadBasic`), `list-joined-teams`, `get-team`, `list-team-channels`, `get-team-channel`, `parse-teams-url`
- auth/account tools

Notably absent: every `list-*-messages`/`get-*-message`/hosted-contents/transcript/recording/presence tool, and destructive management tools (`delete-team-channel`, `remove-team-member`, …).

With #630, `--preset teams-write` + a matching `--allowed-scopes` would need only: `User.Read`, `Chat.ReadBasic`, `ChatMessage.Send`, `ChannelMessage.Send`, `Team.ReadBasic.All`, `Channel.ReadBasic.All` — verified against v0.140.0's `--list-permissions` that this is the minimum viable send-only scope set.

Happy to send a PR (preset tagging is data-only in `endpoints.json`).

nebkat:
Closes #631.

Adds a `teams-write` preset: send in chats, send/reply in channels, name-only chat/team/channel listing, and activity notifications — no message reading, and the universal byte downloaders are deliberately excluded (`omitUniversalUtilities`), since `download-bytes` is an arbitrary-Graph-path reader that would undermine the send-only boundary whenever the token is broader than the preset. A contract test pins the exact tool list, the computed token, and the exclusions so the boundary is CI-enforced.

The second commit makes `Chat.ReadBasic` the primary scope for `list-chats`/`get-chat` and models the Chat hierarchy (`ReadWrite` ⊃ `Read` ⊃ `ReadBasic`) in `SCOPE_HIERARCHY`. Result: `--preset teams-write` alone requests a minimal token (no `--allowed-scopes` incantation), while `--preset teams` collapses the basic tier away — no new consent on upgrade.

Note re #633: the OR-group alternatives introduced there are superseded by the hierarchy — an allowlist containing `Chat.Read` now covers a flat `Chat.ReadBasic` requirement — so these endpoints return to a single flat scope. Narrow filters wanting `lastMessagePreview` add `--extra-scopes Chat.Read`.

Review updates:

- `collapseRedundantScopes` now drops each present lower scope independently, removing the partial-tier gap and the key-ordering dependency (regression-tested with the `^(list-chats|update-chat-message)$` repro).
- The contract test additionally pins `buildAllowedScopeDiagnostics(...).effectivePermissions` to exactly the six write-only Graph scopes (`User.Read` and `offline_access` are injected at the OAuth layer, not endpoint-derived).
- `reply-to-chat-message` removed from the preset (10 tools): the `/chats/{id}/messages/{id}/replies` path family is not a real v1.0 API — verified live, `GET` returns 404 `NotFound: "Requested API is not supported. Please check the path."` (request-id `b1e5540e-a32d-4957-a93e-ec2a8762c922`). Removing the endpoint entirely (it is equally dead in `teams`) is left for a separate PR.
- Open question: `$expand=lastMessagePreview` on `list-chats` under a `Chat.ReadBasic`-only token is assumed gated (the `lastMessagePreview` GET documents `Chat.Read` as its floor) but not yet observed empirically — consent for a `ReadBasic`-only token is pending admin approval on my side. Evidence to be pinned here once available.

820/820 tests, lint/format clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,11 +553,10 @@ npx @softeria/ms-365-mcp-server --preset outlook
 npx @softeria/ms-365-mcp-server --org-mode --preset teams
 ```
 
-The `teams-write` preset is the send-only counterpart to `--read-only`: send/reply in chats and channels, list chats/teams/channels by name, and activity notifications - no message reading and no byte downloaders. Pair it with `--allowed-scopes` so the token cannot read message content either:
+The `teams-write` preset is the send-only counterpart to `--read-only`: send/reply in chats and channels, list chats/teams/channels by name, and activity notifications - no message reading and no byte downloaders. The requested token is minimal by construction (`Chat.ReadBasic`, the `*.Send` scopes, and basic team/channel listing - nothing that can read message content):
 
 ```bash
-npx @softeria/ms-365-mcp-server --org-mode --preset teams-write \
-  --allowed-scopes 'User.Read Chat.ReadBasic ChatMessage.Send ChannelMessage.Send Team.ReadBasic.All Channel.ReadBasic.All TeamsActivity.Send'
+npx @softeria/ms-365-mcp-server --org-mode --preset teams-write
 ```
 
 ## Dynamic Tool Discovery
--- a/README.md
+++ b/README.md
@@ -553,7 +553,7 @@ npx @softeria/ms-365-mcp-server --preset outlook
 npx @softeria/ms-365-mcp-server --org-mode --preset teams
 ```
 
-The `teams-write` preset is the send-only counterpart to `--read-only`: send/reply in chats and channels, list chats/teams/channels by name, and activity notifications - no message reading and no byte downloaders. The requested token is minimal by construction (`Chat.ReadBasic`, the `*.Send` scopes, and basic team/channel listing - nothing that can read message content):
+The `teams-write` preset is the send-only counterpart to `--read-only`: send in chats, send/reply in channels, list chats/teams/channels by name, and activity notifications - no message reading and no byte downloaders. The requested token is minimal by construction (`Chat.ReadBasic`, the `*.Send` scopes, and basic team/channel listing - nothing that can read message content):
 
 ```bash
 npx @softeria/ms-365-mcp-server --org-mode --preset teams-write
--- a/README.md
+++ b/README.md
@@ -539,9 +539,9 @@ npx @softeria/ms-365-mcp-server --preset mail
 npx @softeria/ms-365-mcp-server --list-presets  # See all available presets
 ```
 
-Available presets: `mail`, `calendar`, `files`, `personal`, `work`, `excel`, `contacts`, `tasks`, `onenote`, `search`, `users`, `outlook`, `onedrive`, `teams`, `all`
+Available presets: `mail`, `calendar`, `files`, `personal`, `work`, `excel`, `contacts`, `tasks`, `onenote`, `search`, `users`, `outlook`, `onedrive`, `teams`, `teams-write`, `all`
 
-Each endpoint in `endpoints.json` declares which presets it belongs to via a `presets` array, so every preset is an exact tool-name allow-list that never over-matches across apps (e.g. `mail` does not include shared-mailbox tools; those are in `work`). The universal binary reader `download-bytes` is included in every preset, so whatever an app returns (a file, an attachment, a photo, a recording) can always be fetched; `get-download-url` (a pre-authenticated URL for drive/SharePoint files) rides with the drive-backed presets. So a preset that can find a file can always read its bytes.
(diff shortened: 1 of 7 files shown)


eirikb:
Nice, thanks for this 🙂

I diffed the requested scopes before/after on both branches, and `--preset teams` and `--preset work` come out identical (`Chat.ReadWrite` swallows the rest), so nobody needs to re-consent. Good.

One thing I'd like to be sure about before merging: `list-chats` still takes `$expand`, and Graph documents `$expand=lastMessagePreview` on that operation. The docs kind of sucks, but did you actually see Graph reject it with a `Chat.ReadBasic` only token? You mention it in #633, but I can't find anything in the docs saying that property is gated per permission. If it does come through, the README line about "nothing that can read message content" is a bit too strong. Would be nice to have it pinned somewhere, even just in the PR body.

Two smaller things claude mentioned, both fine as follow-ups:

* `collapseRedundantScopes` only drops the lower scopes when all of them are present (`lowerScopes.every(...)`), so `Chat.ReadWrite` plus `Chat.ReadBasic` without `Chat.Read` keeps both. Presets are fine since `list-pinned-chat-messages` brings in `Chat.Read`, buuut `--enabled-tools '^(list-chats|update-chat-message)$'` requests both. I think deleting each present lower scope on its own would fix it, and it'd get rid of the key ordering dependency too.
* The new test only reads `endpoints.json`, never the computed token. Something pinning `buildAllowedScopeDiagnostics(...).effectivePermissions` to the six scopes would cover the README claim, and I bet it would have caught the one above.

Unrelated, but `reply-to-chat-message` looks like it might not be a real v1.0 operation (replies are only documented for channels). Not your doing, it's already in `teams`, it just stands out more in an 11 tool preset.

nebkat:
Thanks for the close read. All addressed:

**`collapseRedundantScopes`:** fixed as you suggested — each present lower scope is dropped on its own, which also removes the key-ordering dependency. Your `^(list-chats|update-chat-message)$` repro is now a regression test, and it does fail on the old code.

**Computed token:** the contract test now pins `buildAllowedScopeDiagnostics(...).effectivePermissions` for the preset to exactly the six write-only Graph scopes (`User.Read` turns out to be injected at the OAuth layer with `offline_access`, not endpoint-derived). You were right that it would have caught the collapse bug.

**`lastMessagePreview`:** honest answer — no, I haven't observed the rejection myself, and I currently can't: consent for a `Chat.ReadBasic`-only token is stuck behind admin approval in my org. Everything Microsoft documents points to it being gated (`lastMessagePreview`'s own GET lists `Chat.Read` as the floor, and excluding message content is `ReadBasic`'s entire reason to exist), so I've left the README claim standing on that assumption. Any chance you can run `GET /v1.0/chats?$expand=lastMessagePreview` with a `ReadBasic`-only token in your tenant? Happy to pin whatever you observe in the PR body — and if it does come through, I'll strip the claim and we can deal with `$expand` properly.

**`reply-to-chat-message`:** stronger than "might not be" — I verified against live Graph and the path family doesn't exist in v1.0: `GET /chats/{id}/messages/{id}/replies` → 404 `NotFound: "Requested API is not supported. Please check the path."` Dropped it from the preset (10 tools now) and I'll open a follow-up to remove the endpoint entirely, since it's dead in `teams` too.

eirikb:
Thanks, this all looks good. Merging.

On `lastMessagePreview`, I don't have a `Chat.ReadBasic` only token lying around either. I think the safest is to reword the README to what we can actually prove: the token is limited to `Chat.ReadBasic`, which MS defines as chat names and members, not message content. That holds whatever Graph does at runtime, and then we're not claiming the server intercepts anything. I'll try it in my tenant at some point and we can tighten it back up if it turns out to 403.

Small correction: `User.Read` is only injected in the HTTP OAuth path. Stdio goes straight through `resolveAuthScopes`, so `--preset teams-write --login` gets no `User.Read` and `verify-login` will 403 on `/me`. Same for `--preset teams` today, so nothing new, but someone following the new README example will probably hit it first.

PR #644, opened 2026-08-21
feat(audit): record result volume on tool.call audit events

scottjones:
## Problem

The `tool.call` audit event records **which** tool a user called but not **how much** came back. `list-mail-messages` returning 3 items and the same tool returning 30,000 produce identical audit records.

For anyone using the audit log for detection, that is the difference between ordinary use and bulk extraction — and it is currently invisible. `fetchAllPages` widens the gap, since one call can merge up to 100 pages.

The information is already computed a few lines above the `auditLog()` call, but only as human-readable `logger.info` lines with no `request_id`, so it cannot be correlated back to the audit event it belongs to.

## Change

Three optional fields on `AuditEvent`, populated by a `responsePayloadAuditFields()` helper alongside the existing `graphResponseAuditFields()`:

| Field | Meaning |
| --- | --- |
| `result_count` | number of items in a `value` array |
| `result_has_more` | an `@odata.nextLink` was present |
| `response_bytes` | total serialised response size |

**Applied to both audit paths.** The generated Graph tools are the obvious case, but `executeUtilityTool` has its own `auditLog()` call and matters too: `download-bytes` reads raw attachment, recording and profile-photo bytes, and `response_bytes` is the only thing separating a 2KB photo from a 40MB recording. Those tools can return several content items, so `response_bytes` sums all of them rather than the first.

## Design notes

- **Additive only.** The existing `logger.info` lines are untouched, so no current output moves or disappears.
- **Absent rather than zero.** Non-JSON and non-collection responses omit the count fields instead of reporting `0`, so a missing field is never mistaken for an empty result.
- **`result_has_more` is correct after a merge.** `fetchAllPages` deletes `@odata.nextLink` when it merges, so a fully-merged response reports no more pages.
- **Follows existing precedent.** `AuditEvent` already carries `graph_batch_subrequest_count`, so count fields and the `snake_case` naming are consistent.

## Known limitation

`result_count` and `result_has_more` require JSON. Under `--toon` the payload is TOON, so only `response_bytes` is recorded. Documented in place.

Fixing it properly means deriving the count before serialisation — but `fetchAllPages` merges and re-serialises in `graph-tools` *after* the client returns, so a client-side count would be wrong for merged responses. Happy to take that on if you have a view on where the computation belongs; I did not want to guess at a restructure.

## Testing

Four tests in `src/__tests__/graph-tools.test.ts`, following the shape of the existing `graph_batch_*` audit assertions:

- collection response records count, pagination and size
- single-object response records size but omits `result_count` / `result_has_more`
- non-JSON response (`get-mail-message-mime`) still records size
- `response_bytes` sums every content item, not just the first

Full `verify` gate passes locally: `tsc --noEmit`, `eslint` (0 errors), `prettier --check`, `tsup` build, and **904 tests across 66 files** (900 on main).

## Context

We self-host this server for internal Outlook access, with a tool allowlist and scope ceiling as the security boundary. Our security team asked what the audit trail would show if someone extracted a mailbox in bulk, and the honest answer was "the same as reading three emails". We would much rather consume a released version than carry a patch.


eocampo3030 (commented):
Verified locally: checked out `feat/audit-result-volume` and ran the full `verify` gate — `generate`, `eslint`, `prettier --check`, `tsc --noEmit`, `tsup`, **904 tests / 66 files, all passing**. The claims about the gate hold up.

The premise is right and the implementation is careful — additive-only, absent-not-zero, and it correctly covers the utility path that a narrower patch would have missed. Everything below is about whether the fields deliver the detection value the PR is aiming for.

### 1. `download-bytes-to-file` reports ~150 `response_bytes` for an arbitrarily large extraction (highest-value finding)

The tool built specifically for out-of-band bulk saves returns a small envelope — `{ path, contentType, bytesWritten }` ([graph-tools.ts#L827-L836](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-tools.ts#L827-L836)) — so `response_bytes` measures the receipt, not the payload. A 250MB mail attachment written to disk audits at roughly the same size as an error message. A detection rule thresholding on `response_bytes` misses the single most exfil-shaped tool in the server, which is exactly the gap this PR sets out to close.

The real number is already sitting in the payload the helper just parsed: `bytesWritten: result.contentLength`. Reading it (or `contentLength`, which `download-bytes` also returns) would close this in a few lines.

Related, and worth a line in the doc comment so operators don't over-trust the field: `get-download-url` hands back a pre-authenticated URL, so the bytes never traverse the server at all — inherently unobservable, not this PR's problem, but it means `response_bytes` is a floor, not a ceiling.

### 2. New full `JSON.parse` on the utility path, where there was none

`executeUtilityTool` ([graph-tools.ts#L330-L352](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-tools.ts#L330-L352)) never touched the payload text before this change. Now every utility call parses it — including `download-bytes`, whose response is a base64 blob. Measured on a 40MB attachment (53MB JSON envelope): **+19ms and a transient ~54MB heap spike**, to discover there is no `value` array.

This matters more than the raw numbers suggest, because the codebase deliberately avoids exactly this: [graph-client.ts#L289](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-client.ts#L289) explains that `download-bytes-to-file` exists to keep big payloads out of memory. A size guard, or skipping the parse when the text cannot be a collection, would avoid it.

### 3. The graph path now parses the same text twice

[graph-tools.ts#L1721-L1735](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-tools.ts#L1721-L1735) already parses `responseText` and pulls out both the item count and the `nextLink` for the `logger.info` lines, immediately above the `auditLog()` call. The new helper parses it again. On an 8MB merged response (10k items, the `DEFAULT_MAX_ITEMS` ceiling) that is ~12ms and a duplicate object graph, for numbers already in scope.

Reusing that parse keeps the `logger.info` output byte-identical — the additive-only goal is untouched, only the source of the numbers changes.

### 4. Answering the open design question: `_meta` is the right home, and the stated blocker is small

The PR asks where the pre-serialisation computation belongs, and there is already an exact precedent. `graph_batch_subrequest_count` is derived from the in-memory parsed object at [graph-client.ts#L260-L262](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-client.ts#L260-L262) and carried on `_meta`, which `graphResponseAuditFields` then lifts out. Adding `result_count` / `result_has_more` beside `extractBatchMetadata(result)` costs nothing (the object is already parsed), is format-agnostic — **which also fixes the `--toon` limitation** — and removes both parses above.

The blocker as described ("a client-side count would be wrong for merged responses") looks genuinely solvable: the merge block already holds `allItems` and already deletes `@odata.nextLink` at [graph-tools.ts#L1696-L1701](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-tools.ts#L1696-L1701), so it can overwrite `response._meta` with the merged count and `has_more: false` right there. Same for `download-bytes`: `contentLength: buffer.byteLength` is computed at [graph-client.ts#L237](https://github.com/Softeria/ms-365-mcp-server/blob/8742eebdd483f583241dcef63782b239bf7037ee/src/graph-client.ts#L237), so the true byte count is available pre-base64 — which fixes finding 1 and the ~1.37x base64 inflation in one move.

### 5. `result_has_more` should be an explicit `false` when the payload parsed as a collection

The PR's absent-not-zero reasoning argues for this rather than against it. When `Array.isArray(parsed.value)` holds, both answers are known — so emitting `false` lets a query distinguish "complete result" from "unknown / not JSON". As written those two cases are the same absence, and only the incidental presence of `result_count` lets a consumer tell them apart.

### 6. `result_count` under-reports for nested collections

Microsoft Search's `search-query` returns `value: [{ hitsContainers: [{ hits: [...] }] }]` — 500 hits audit as `result_count: 1`. The semantic `retrieval` tool returns `retrievalHits: [...]` with no top-level `value`, so no count at all. Both are bulk-read tools. Batch is fine, covered by `graph_batch_subrequest_count`. Worth naming in the doc comment so nobody thresholds on `result_count` for search.

### Nits

- **Asymmetry, and a test for a shape that cannot occur.** `response_bytes` sums every content item but the JSON parse reads only `texts[0]`. Checking every `content: [...]` literal in `src/`, nothing produces more than one text item — so "those tools can return several content items" does not hold today, and the summing test covers a hypothetical. Harmless defensiveness; just worth not letting the two fields disagree about what "the response" is if a second item ever does appear.
- **`docs/deployment.md` L216** lists the audit record's fields and states that returned content and Graph response bodies are never recorded. The three new fields belong in that list, with a word that response *metadata* (size, count) is now included — compliance readers lean on that sentence.

### Verdict

Worth merging with changes. Findings 1 and 4 are the ones I would want addressed first: as it stands the field set has a blind spot on precisely the highest-volume tool, and the `_meta` route resolves that, the `--toon` limitation, and both redundant parses together — using the pattern the file already established for batch counts. Findings 5, 6 and the docs nit are quick. The test shapes follow the existing `graph_batch_*` conventions well and I would keep them.

eocampo3030 (changes requested):
Formalising the verdict from my [detailed review above](https://github.com/Softeria/ms-365-mcp-server/pull/644#pullrequestreview-4994734601) as a request for changes. The direction is right and I would like to see this land — the blocking items are findings 1 and 4:

1. **`download-bytes-to-file` reports ~150 `response_bytes` for an arbitrarily large extraction.** The field set has a blind spot on precisely the most exfil-shaped tool in the server, which is the gap the PR sets out to close. The real number (`bytesWritten` / `contentLength`) is already in the payload the helper parses.
2. **Derive the counts in `graph-client` and carry them on `_meta`,** beside the existing `extractBatchMetadata(result)` call. Same precedent as `graph_batch_subrequest_count`, costs nothing because the object is already parsed, fixes the `--toon` limitation, and removes both the new utility-path parse (finding 2) and the duplicate graph-path parse (finding 3). The `fetchAllPages` concern raised in the description is a couple of lines at the merge site, which already holds `allItems` and already deletes `@odata.nextLink`.

Findings 5, 6 and the two nits are non-blocking and quick. Happy to re-review as soon as you have pushed.

scottjones:
Thank you — this is a substantially better patch for it. All six findings addressed; pushed.

**Finding 4 — `_meta` is the right home, and you were right that the blocker was small.** Taken wholesale. Counts are now derived in `extractPayloadMetadata()` beside `extractBatchMetadata()` and carried on `_meta`, which `graphResponseAuditFields` lifts. That removes the utility-path parse (2) and the duplicate graph-path parse (3), and fixes `--toon` for free — the field set no longer depends on the serialisation format at all. My "wrong for merged responses" objection was indeed a couple of lines: the merge site restates the count and sets `result_has_more: false` right where it already deletes `@odata.nextLink`.

**Finding 1 — fixed, and it was the worst of them.** `download-bytes-to-file` now carries `response_bytes: result.contentLength` on `_meta`, so it reports the file written rather than the receipt. Your framing was the useful part: a field set aimed at spotting bulk extraction that was blind on the one tool built for bulk extraction. Binary content also now uses the pre-base64 `buffer.byteLength`, so the ~1.37x inflation is gone too — one change, both problems, exactly as you said.

**Finding 5 — agreed, and you turned my own argument around on me.** `result_has_more` is now explicitly `false` for a complete collection. "Complete" and "not a collection" were the same absence, which is precisely what absent-not-zero is supposed to prevent.

**Finding 6 — documented rather than fixed.** The doc comment now names Microsoft Search's `value: [{ hitsContainers: [{ hits }] }]` and warns against thresholding `result_count` for search tools. Counting nested shapes generically felt like guessing at each tool's semantics; happy to special-case search if you would rather.

**Nits.** You were right that nothing produces more than one text item — the summing behaviour and its test are gone, so the asymmetry no longer exists. `docs/deployment.md` L216 now lists the three fields and distinguishes response *metadata* from response *content*, keeping the "never recorded" claim true for compliance readers. I have also noted in the comment that `get-download-url` bytes never traverse the server, so `response_bytes` is a floor.

Tests restructured to follow the new seam: four in `graph-client.test.ts` covering derivation (collection, complete-collection `false`, single object, pre-base64 binary), three in `graph-tools.test.ts` covering the `_meta` lift including that `false` survives. **910 tests / 66 files**, full `verify` gate green.

Two things I did not do, both flagged for your call rather than assumed:
- **`get-download-url`** — inherently unobservable, as you say. Not touched.
- **`retrievalHits`** — no top-level `value`, so no count. Same reasoning as finding 6.

Ready for re-review whenever suits.

eirikb:
This looks good now, thanks for the rework - `_meta` was the right seam.

It conflicts with main now though. Could you merge main in (not rebase, #647 got squashed)? `audit-log.ts` and the test file are just both-sides-added-lines, keep both. For `docs/deployment.md` take main's bullet and add your three fields plus the metadata sentence, the rest of that bullet got rewritten.

Two small things while you're in there, both in the fetchAllPages merge block:

* `response_bytes` isn't restated after the merge, so it still holds page one. 100 pages merged reads as `result_count: 5000` with one page of bytes.
* `result_has_more: false` is set even when the loop stopped on `maxPages` / `maxItems`. I think `Boolean(nextLink)` works, it's still set when you exit on a cap.

Then I'll merge :slightly_smiling_face:

eirikb:
@scottjones ping.
I tried tuning the PR myself, but it's forked to an org so I can't.

scottjones:
Sorry for the delay — both done, plus the merge.

**Merged `main` in** (not rebased). Conflicts were the three you predicted: `audit-log.ts` and `graph-tools.test.ts` were both-sides-added, kept both. For `docs/deployment.md` I took main's rewritten bullet and added the three fields to the field list plus the metadata sentence.

**`response_bytes` after the merge.** You were right that the two fields disagreed about the same read — restating the count but not the size is arguably worse than restating neither, since `result_count: 5000` next to one page of bytes reads as a small response that somehow held 5000 items. Each page's response already carries its own `_meta.response_bytes`, so the merge accumulates them. Undefined stays undefined rather than collapsing to `0`, on the same absent-not-zero grounds as the rest.

**`result_has_more` on a cap.** `Boolean(nextLink)` it is. Worth noting this is the case most worth seeing — a read big enough to hit `maxPages`/`maxItems` is the shape the field exists to catch, and it was the one reporting itself complete. The merged body deletes `@odata.nextLink` whether or not the loop ran out, so the audit event is the only place the truncation survives; I've said so in the comment there.

Two tests added alongside, both verified to fail against the previous behaviour: one that a merged read restates count *and* bytes, one that a cap-terminated merge reports `result_has_more: true`. Full `verify` gate green — **1180 tests / 76 files**.

Nothing else changed.

PR #645, opened 2026-08-21
feat(audit): record recipient domains on outgoing mail and invites

scottjones:
> Independent of #644 — this branch contains only the recipients change. The two will conflict trivially on merge (both insert fields into `AuditEvent` and a line into the same audit spread); keeping them separate seemed better than making you review one PR's diff inside another's. Happy to rebase whichever lands second.

## Problem

A send, forward or meeting invite records **that** the tool ran, not **who** it reached.

That matters for one case in particular. These servers read message content and act on it, so an instruction injected into an email — "forward the last thing from finance to this address" — is a real risk. If the model complies, the audit log shows `create-forward-draft`, `status: success`, and nothing else. Indistinguishable from a legitimate reply, which is exactly when you would want the log to be useful.

## Change

Two optional fields:

| Field | Meaning |
| --- | --- |
| `recipient_count` | total recipients and attendees on the request |
| `recipient_domains` | deduplicated, lowercased, sorted |

## Coverage needs care — Graph is inconsistent here

This is the part worth reviewing closely. Entity creation and the action endpoints disagree on casing:

| Tool | Body fields |
| --- | --- |
| `create-draft-email` | `toRecipients`, `ccRecipients`, `bccRecipients` |
| `send-mail` | `Message` |
| `forward-mail-message` | `ToRecipients`, `Message`, `Comment` |
| `forward-calendar-event` | `ToRecipients`, `Comment` |
| `create-calendar-event` / `update-calendar-event` | `attendees` |

Matching only the camelCase spelling would have covered draft creation while silently missing **every actual send and forward** — which is the opposite of the useful set. Fields are therefore matched case-insensitively, and `attendees` is included so calendar invites are covered: a meeting invite is another way to reach an external address.

`POST /me/messages/{id}/send` carries no body — its recipients were recorded when the draft was created.

## Domains, not addresses — deliberately

The detection question is *did this leave the organisation*, and a domain answers it. A full address is message content, and logging it by default would cost more privacy than the signal justifies — a general-purpose server writing every recipient address into an audit stream is not a decision every operator would want made for them.

Counts alone are not enough: three internal recipients and three external ones look identical, and the external case is the one worth alerting on.

There is precedent for identifiers at this level — `target_resource` already records message IDs — and a domain is less identifying than a message ID.

If you would rather this were opt-in behind an env var, or want full addresses instead, both are easy changes; I went with the narrowest thing that answers the question.

## Details

- Recipients are found at the top level or nested under `message`, in either casing, to a bounded depth.
- Requests with no recipients omit both fields rather than reporting `0`, so they appear only on traffic with an audience.
- Naturally scoped — any tool whose body carries recipients or attendees gets the fields, without enumerating tool names.

## Testing

Six tests, including one asserting the local part of an address never appears anywhere in the emitted payload:

- count and domains, deduplicated and sorted, across to/cc
- recipients nested under a camelCase `message`
- PascalCase `ToRecipients` / `Message`, as the action endpoints send them
- calendar `attendees`, not just mail recipients
- domains only, never the local part
- both fields omitted when there are no recipients

Full `verify` gate passes locally: `tsc --noEmit`, `eslint` (0 errors), `prettier --check`, `tsup` build, and **906 tests across 66 files** (900 on main).

## Context

We self-host this for internal Outlook access with a tool allowlist as the security boundary — read plus draft-creation, no send and no delete. Our security review asked what the audit trail would show if an injected instruction addressed a draft outside the company, and the honest answer was "nothing distinguishable". This closes that.

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -409,6 +409,115 @@ describe('graph-tools', () => {
     });
   });
 
+  describe('audit recipient metadata', () => {
+    const draftEndpoint = () => {
+      const endpoint = makeEndpoint({
+        method: 'post',
+        path: '/me/messages',
+        alias: 'create-draft-email',
+        parameters: [{ name: 'body', type: 'Body', schema: z.object({}).passthrough() }],
+      });
+      const config = makeConfig({
+        pathPattern: '/me/messages',
+        method: 'post',
+        toolName: 'create-draft-email',
+      });
+      mockEndpoints.push(endpoint);
+      mockEndpointsJson = [config];
+    };
+
+    const runDraft = async (body: unknown) => {
+      const graphClient = createMockGraphClient([
+        {
+          content: [{ type: 'text', text: JSON.stringify({ id: 'draft-1' }) }],
+          _meta: { http_status: 201 },
+        },
+      ]);
+      const server = createMockServer();
+      const { registerGraphTools } = await loadModule();
+      registerGraphTools(
+        server as unknown as Parameters<typeof registerGraphTools>[0],
+        graphClient as unknown as Parameters<typeof registerGraphTools>[1]
+      );
+      await server.tools.get('create-draft-email')!.handler({ body });
+      return auditLogMock.mock.calls[0][0];
+    };
+
+    it('records recipient count and domains, deduplicated and sorted', async () => {
+      draftEndpoint();
+      const payload = await runDraft({
+        toRecipients: [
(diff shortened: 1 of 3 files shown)


eirikb:
Hi there, and sorry for the late reply.

CI is green, merging.

Two things I can fix myself after merge: `result_has_more` is hardcoded `false` after the page merge even when the loop stopped on the cap, so a truncated 10k read logs as complete if I read it right. And `response_bytes` still only covers page one. Few lines each, same spot.

Issue #569, opened 2026-07-17
create-draft-email (POST /me/messages) drops subject and toRecipients — only body field reaches Graph, causing UnableToDeserializePostBody


oleksandrilyin:
Version: 0.131.1 (via `npx @softeria/ms-365-mcp-server@latest`)

Calling `create-draft-email` with a full set of arguments (`subject`, `body: {contentType, content}`, `toRecipients: [...]`) fails with:
```
Microsoft Graph API error: 400 Bad Request - {"error":{"code":"UnableToDeserializePostBody","message":"were unable to deserialize "}}
```

Confirmed via the server's own `mcp-server.log` (`Calling /me/messages with options: ...`) that the actual outgoing POST body was **only** `{"contentType":"text","content":"..."}` — `subject` and `toRecipients` never reached the request at all, silently dropped before serialization. Looks like the input schema for this tool only maps the `body` sub-object, not the full `microsoft.graph.message` resource, unlike `send-mail` which correctly nests everything under a `message`/`body` wrapper.

**Repro:** call `create-draft-email` with `{subject, body: {contentType, content}, toRecipients: [{emailAddress:{address}}]}` — 400 every time, any content, including plain ASCII.

**Workaround confirmed working:** same payload sent as a raw `POST /me/messages` sub-request via `graph-batch` — succeeds (201, valid draft, correct subject/body/recipient).

**Ask:** map `create-draft-email`'s input schema to the same shape/handling as `send-mail`, so `subject`/`toRecipients`/`body` all reach the Graph payload.

eirikb:
Uhm should have been fixed in #570

PR #655, opened 2026-08-31
feat(tools): cap served tool-definition size to keep every tool listed

okapi-ca:
## Problem

Some MCP clients serve a **bounded total amount of tool-definition data per connector**. The tool list is filled **alphabetically**, each tool is charged on admission, and a definition that no longer fits is **skipped entirely** rather than truncated — so smaller definitions sorting after the gap still make it through, while large ones vanish.

With the full tool surface enabled this server is roughly 1.7× such a limit, and the tools that disappear are mostly the write tools: `send-*`, `reply-*`, `update-*`. They carry request bodies, bodies carry the bytes, and they happen to sit late in the alphabet. The stdio/CLI path is unaffected, which is why this is easy to miss.

## The rule

Confirmed by Anthropic support on 2026-09-01, in reply to our ticket (nothing public documents it, which is why I am relaying it here rather than linking a doc):

- **1,048,576 bytes per connector**, walked **strictly alphabetically**, charging each tool on admission and skipping any that no longer fits.
- Counted in **bytes, not tokens**, and **only the input schema counts** — the UTF-8 length of `inputSchema` serialized compactly, with `name` and `description` excluded.
- **Per connector.** One connector's tools cannot be evicted by another's, so splitting a toolset across two connectors is a real mitigation.
- The 256 aggregate tool count is no longer the cap.

The falsification test from the review thread also came back: holding the tool count fixed and changing only definition size (this branch, deployed to the same connector on Aug 31) brought every tool back. Full write-up in the thread below.

## What this PR does

Two things, separable:

1. **Measures** what `tools/list` serves — every tool's input schema, serialized exactly as the transport does (the SDK's conversion, then `normalizeToolSchemaRefs`) — and logs the total against the 1 MiB connector budget once per surface. Always on. Until now nothing told an operator they had just served 335 tools and 1.8 MB of schema.
2. **Optionally prunes** oversized request-body schemas to a per-tool ration. **Opt-in**: `MS365_MCP_TOOL_DEFINITION_BUDGET=<bytes>`. Unset, every tool registers byte for byte as generated.

## Measurement

Through the real `tools/list` handler on this branch (rebased on `main` at 41d766a, client regenerated 2026-09-08, 335 tools, org mode, writes enabled, HTTP mode):

|                                           | Unrationed         | `MS365_MCP_TOOL_DEFINITION_BUDGET=4096` |
| ----------------------------------------- | ------------------ | --------------------------------------- |
| Schema bytes served                       | 1,831,221 B        | **688,812 B (−62%)**                    |
| Share of the 1 MiB connector budget       | 174.6%             | **65.7%**                               |
| Tools skipped by the alphabetical walk    | **123 of 335**     | **0**                                   |
| Tools pruned                              | —                  | 82 of 335                               |
| Largest schema (`create-sharepoint-list`) | 84,167 B           | 3,447 B                                 |
| Median schema                             | 2,082 B            | 2,082 B                                 |
| Registration, first / subsequent          | 74 ms / 9 ms       | 188 ms / 9 ms                           |

The logged aggregate equals this probe to the byte in both modes — the measurement runs the same conversion and the same ref normalization the transport does, so it cannot drift from what is served. Since #643 inlines `$ref`s back wherever a client might not resolve `$defs`, the wire form is larger than a refs-kept estimate; the aggregate reflects the post-#643 form.

The weight is not in `endpoints.json` but in the Graph OpenAPI body shapes the generated client inlines: `create-sharepoint-list` occupies 670 bytes of endpoint config and expands to ~84 KB of served schema. The distribution is extremely skewed, and 253 of 335 tools are under the 4 KiB ration untouched.

## Approach (when opted in)

Fit each definition under the ration by pruning **body** schemas, and **only for tools that actually overflow**. Pruning walks an ordered list of strategies, least to most destructive:

| Order | maxDepth | describeDepth |
| ----- | -------- | ------------- |
| 1     | 4        | 2             |
| 2     | 3        | 2             |
| 3     | 2        | 2             |
| 4     | 1        | 2             |
| 5     | 1        | 1             |
| 6     | 1        | 0             |
| 7     | 0        | 0             |

The ordering is the point. Field *names* are what a model needs to compose a request; the per-field *description* text is what actually consumes the budget. Trimming descriptions before structure keeps every body's field names visible — **bodies made opaque by the ration: zero**; on this surface the 82 pruned bodies land between strategies 1 and 5, none reaches 6 or 7. The eight bodies that show no field names are opaque at any budget, unlimited included; that is inherent to the spec (raw uploads, `@odata.id` references, free-form batch bodies).

Why this is safe:

- Graph validates bodies server-side, and bodies already go through `lenientBodySchema`.
- `passthrough()` is preserved, so flattened bodies still reach the handler — the #569 regression has a dedicated test here.
- `optional` / `nullable` / `default` wrappers are rebuilt around the pruned inner type. Turning an optional field required would reject calls that are valid today.
- `z.union` alternatives are pruned at the same depth (choosing between them is not a step deeper); `z.record` values one level down, like array elements. The generated client wraps most Graph object references as `z.union([named, z.object({}).partial().passthrough()])`, so a union is the common shape of a nested body field.
- Collapsed values keep their kind (object → record, array → array), so the expected shape stays visible.
- Recursive `lazy` Graph types (driveItem → children → driveItem) terminate on the depth budget.

Pruning costs schema-level guidance, not the ability to make the call. It also cuts definition tokens for every client, stdio included.

## Cost

HTTP mode is stateless and rebuilds the `McpServer` — and so re-registers every tool — on each `POST /mcp`. Fit results are memoised per tool and per configuration fingerprint, so the measuring pass runs once per process: the first request pays 74 ms unrationed / 188 ms rationed on 335 tools, every request after that 9 ms.

## Observability

Three log lines, each learned the hard way:

- The **aggregate** input-schema bytes as served by `tools/list`, against the 1 MiB connector budget, once per surface. When the total crosses the ceiling it warns and names the ways out: set a ration, `--discovery`, narrow `--enabled-tools`, split across two connectors.
- When a ration is set, a body that loses its field names is logged at `warn` on its own, separately from ordinary depth trimming.
- A tool that still overflows at the last strategy is registered and logged rather than dropped.

## Verification

`npm run verify` on this branch: 0 lint errors, Prettier clean, **1050 tests / 72 files passing**, including a test that pins the measurement to the `tools/list` wire form byte for byte against a real `McpServer`, and coverage of the strategy ordering, depth pruning, union/record/default handling, optionality and nullability preservation, passthrough, recursive termination, minimal-pruning selection, non-body params left untouched, the fit cache, and the env-var semantics (unset → no pruning; invalid → recommended ration). `src/endpoints.json` is untouched. README documents the variable.

Two commits on top of `main` at 41d766a: the original change, then the review follow-up (opt-in, exact measurement, once-per-process, union handling). Happy to squash before merge.


okapi-ca:
## Correction to the PR description

One claim in the description above is wrong, and I would rather flag it myself than have you find it in review.

I wrote that "top-level property names survive at depth 1, which is what a model needs to compose a request." That is true of the mechanism but **not true of the result** at the 4 KiB default. For the largest bodies, depth 1 still does not fit the budget, so the loop falls through to depth 0 and collapses the whole body into an open record — the model then sees no field names at all.

I had only validated total served volume, never what the budget produced per tool. Measuring it properly (diffing the served `inputSchema` with the budget applied against the budget raised out of the way) gives:

- **248 of 327 tools: byte-for-byte unchanged** — already under budget.
- **42 tools keep their top-level body field names**, losing only deeper structure. `send-mail` keeps 2/2, `create-todo-task` keeps 21/21. This is the outcome I described.
- **37 tools lose all body field visibility.** Notably `send-chat-message`, `send-channel-message`, `create-calendar-event`, `update-mail-message`, `create-team-channel`, `create-chat`, `update-todo-task`, `create-planner-task`.

Execution is unaffected either way — the handler forwards the body and Graph validates it — but for those 37 a model no longer sees that `create-calendar-event` takes `subject` / `start` / `end` / `attendees`. That is a real quality regression, and a silent one: it shows up as malformed calls, not as an error.

## The tradeoff, measured

| Default budget | Total served | Bodies made opaque |
|---|---|---|
| 4 KiB (this PR) | 791,392 B | 37 |
| 6 KiB | 914,044 B | 12 |
| 8 KiB | 1,036,963 B | 2 |

Eight bodies are opaque regardless of the budget — that is inherent to the spec, not caused by pruning.

## Options

1. **Raise the default to 6–8 KiB.** 8 KiB nearly eliminates the regression (2 tools) at a 47% reduction, but lands above the limit if it is really half of the unpruned volume. 6 KiB is the safe middle.
2. **Trim descriptions before collapsing structure.** Field *names* are what a model needs; the verbose per-field descriptions are what actually blow the budget. Dropping nested descriptions first would keep names visible even at 4 KiB. This is the correct fix rather than a knob, and I am happy to implement it in this PR.
3. **Ship it opt-in**, default off, so nothing changes for existing users unless they set the env var.

My preference is option 2, with option 1 as the fallback if you would rather keep the change minimal. Let me know which you want and I will push it — I would not merge this as-is now that the per-tool cost is known.

okapi-ca:
Pushed the fix from option 2 and updated the PR description to match.

Pruning now trims nested descriptions before structural depth, and the body itself only as a last resort. At the same 4 KiB default, bodies made opaque by the budget go from **37 to 0** — `create-calendar-event`, `send-chat-message`, `send-channel-message` and the rest keep their field names. Total served volume is 850,544 B instead of 791,392 B, so 59 KB buys back the field names of 37 tools: −57% instead of −59.5%.

The eight bodies that still show no field names are opaque at any budget, unlimited included — spec-inherent, not caused by pruning.

That also makes the option-1 budget bump moot: at 6 KiB the count is still those same eight, so raising it would spend headroom against the connector limit for no discoverability gain. Left the default at 4 KiB.

Also added a dedicated `warn` for a body that loses its field names, separate from ordinary depth trimming, so that particular regression cannot recur silently.

`npm run verify`: 1025 tests / 69 files green. Branch is a single commit. Still happy to make it opt-in if you prefer — just say so.

eirikb:
Hi @okapi-ca

Thanks for the measurements, I checked them against the generated client here and they hold up. The body shapes really are that heavy, and I appreciate you catching your own claim and re-measuring instead of leaving it standing.

But I'm not 100% sure about the mechanism. What I can find documented is a cap on the *number* of tools (256 aggregate across connectors), not on bytes. And counting what we actually register: org mode with writes is 330, which is the only config that goes over. Personal is 182, read-only 74, org + read-only 164.

A plain count cap would drop the same tools you're seeing go missing. Sorted alphabetically and cut at 256, every `update-*`, `send-*` and `reply-*` falls off the end, 43 of 43. And your own table has tools registered going 333 -> 333, so if it is the count then pruning bytes wouldn't move anything.

Could you try `--preset work` (156 tools) or `--discovery` on the setup where the write tools disappear? If they come back it's the count, and I think the right fix is then just a warning when the surface goes over ~256, pointing at the presets. If they're still missing then you're right about the bytes and I'll take the PR.

The other thing that would settle it: a raw `tools/list` from that client showing a small tool arriving after a big one went missing. Under a count cap nothing after the cut arrives at all.

Sorry to ask for more measuring when you already did a lot 😅

okapi-ca:
No apology needed — it's the right question, and the data to settle it was already in hand. It's size, not count: the connector went past 256.

**269 tools arrived, not 256.** Our deployment of this fork serves 327 tools over HTTP in org mode with writes. The org connector's *Tool permission restrictions* page enumerated 269 of them — corroborated name for name by a Claude Desktop session on the same connector, which also returned 269 when asked for an exact count rather than an estimate. A 256 cap cannot produce 269.

**14 of those 269 sort after alphabetical position 256, including the very last name in the list.** Under a count cap with an alphabetical cut, nothing after the cut arrives at all. Re-probing `tools/list` with the budget lifted (org, writes, HTTP — 327 tools, 1,882,455 B on our endpoint set) and lining the sizes up against what the connector actually enumerated:

| tool | served bytes | alpha pos | connector |
|---|---|---|---|
| `update-team-channel` | 75,517 | 322 | missing |
| `update-sharepoint-list-item` | 47,873 | 318 | missing |
| `send-chat-message` | 18,849 | 278 | missing |
| `search-query` | 9,737 | 275 | missing |
| `send-mail` | 9,045 | 280 | missing |
| `upload-file-content` | 1,881 | 325 | missing |
| `upload-my-profile-photo` | 1,821 | 326 | **served** |

That last pair is the artifact you asked for: alphabetically adjacent, ~60 bytes apart, the smaller one arrives — and it arrives after fifty-odd larger tools were skipped. In aggregate the omitted median 8,207 B against 2,732 B for the served.

Two of the missing are read tools (`search-query`, `search-sharepoint-sites`), lost at the tail of the page's read-only section (161 of 163), while the write section showed a hole in the middle with its last element present. One global count cap doesn't produce two separate cuts in two independently sorted sections.

Fitting a one-parameter model — walk alphabetically, skip what doesn't fit, keep going — a total budget of ~1.30 MB reproduces the observed omission set at 0.95 Jaccard: 59 predicted against 58 observed, three name-level disagreements out of 327. One wrinkle I can't explain in bytes: `unpin-chat-message` (1,701 B, pos 293) is missing while `upload-my-profile-photo` (1,821 B, pos 326) arrives. No monotone byte threshold does that, so the accounting unit is likely tokens rather than bytes — description text tokenizes very unevenly. Size-based either way.

**And your own falsification test came back.** You're right that under a count cap, pruning bytes would move nothing — 333 registered before, 333 after. We deployed this branch's pruning to that same connector on Aug 31: identical tool count, only definition size reduced. Confirmed today, claude.ai enumerates all 327. Nothing else changed, and the connector was not recreated — we had already established that recreating it changes nothing.

On `--preset work`: I don't think it discriminates. 156 tools is below 256 *and* below any plausible size budget, so both hypotheses predict the tools come back. The experiment that separates them is the one above — hold the count fixed, change only the size — which is what this PR does.

Also pushed: the module header still carried the "top-level property names survive at depth 1" claim I retracted upthread. Replaced with what the shipped strategy order actually guarantees. Comment only, no behavior change, branch still a single commit.

Still happy to put it behind an env var and ship it opt-in if you'd rather keep the default untouched.

okapi-ca:
Support came back this morning and settled it. The rule, as given to us:

- The budget is counted in **bytes, not tokens**, and only the tool's **input schema** counts — the UTF-8 length of the schema serialized compactly, with `name` and `description` excluded.
- **1,048,576 bytes per connector**, walked **strictly alphabetically**, charging each tool on admission and skipping any that no longer fits.
- The budget is **per connector**. One connector's tools cannot be evicted by another's, so splitting a toolset across two connectors is a real mitigation rather than a guess.

And on your question directly: *"the 256 aggregate tool count your maintainer found is no longer the cap."* Nothing public documents either limit today, which is why I am relaying it here rather than linking a doc.

So the mechanism is the one this PR is built on. But the same rule catches an error in the PR, and it is mine.

`measureToolDefinitionBytes` folds `name` and `description` into the count. The platform charges for neither. The ration has therefore been pruning request bodies to pay for text that is free — over-pruning, in the one direction that costs the model schema guidance.

Corrected to measure the input schema alone. Over our fork's 327-tool org surface: 78 tools pruned instead of 79, and **14,390 bytes of schema handed back** to bodies that did not have to lose it. Total served goes 635,399 → 649,789 bytes, 62% of the megabyte, nothing skipped by the alphabetical walk. These are schema-only figures, so they are not comparable to the 850,544 B I quoted earlier, which counted the whole definition.

Two smaller consequences worth having:

- A verbose `llmTip` costs nothing against this budget. It still costs model context, which is a separate constraint — but the two are no longer the same trade-off.
- It explains the one anomaly I could not account for last time: `unpin-chat-message` (1,701 B) was omitted while `upload-my-profile-photo` (1,821 B), 33 positions later, was served. Measured on the whole definition `unpin-chat-message` is the smaller of the pair; measured on the schema alone it is the larger. The unit was never the problem — my measurement's scope was.

I would also add an aggregate log line at registration. The per-tool ration is a means; the total is what actually decides whether a tool is served, and right now nothing tells an operator where they stand against a ceiling that is documented nowhere.

Both changes are committed on our side. Say the word and I will push them onto this branch.

okapi-ca:
Correction to my comment above, before you act on it: three of the numbers in it are wrong, and wrong in the direction that overstates the problem.

I measured by re-serializing each tool's schema locally with `$refStrategy: 'none'`. That is not what goes out. `tools/list` emits the SDK's own conversion, then `normalizeToolSchemaRefs` on top, and both keep `$ref`/`$defs`. Inlining every shared subtree over-states ref-heavy Graph bodies by an order of magnitude. Re-measured through a real `tools/list` probe over the same 327-tool org surface, unrationed:

| | I said | Actually served |
|---|---|---|
| Total schema bytes, unrationed | 4,954,115 | **1,666,197** (1.6× the budget, not 4.7×) |
| Tools skipped by the alphabetical walk | 259 of 327 | **78 of 327** |
| `create-sharepoint-list` alone | 1,000,754 B | **83,965 B** |

Your original 103,368 B figure for that tool was the honest one — measured through the transport, on the whole definition. Mine was an artifact of the measurement.

What does not change: the capped figures. 634,263 → 648,653 bytes served with the schema-only fix, 61.9% of the budget, nothing skipped — those came from the same real probe and hold. The rule from support holds too, and 1.6× the budget is still comfortably over it, which is what this PR exists for.

There is a consequence for the PR itself, though, and it is more interesting than the arithmetic. `measureToolDefinitionBytes` rations against the inlined form, so it charges tools for bytes that are never sent — and it does that hardest on exactly the ref-heavy write bodies that carry the volume. It errs safe (it never under-prunes) but it prunes more than the budget requires, which costs the model schema guidance for no gain.

I have not changed that here. Aligning the measurement with the served form means running the SDK conversion plus `normalizeToolSchemaRefs` at registration time, and it would reshape which tools get pruned and how far — a behavioural change that deserves its own diff and its own numbers rather than being folded into this one. Say which you would rather review: this PR as it stands with the schema-only fix, or with the ref-accurate measurement in the same change.

Apologies for the noise. I would rather flag my own bad numbers than have you check them against your generated client and find them yourself, which is what you did with my last claim.

eirikb:
Hi @okapi-ca

Sorry, took a while. You were right about the bytes. Measured it here too, serialising the way `tools/list` actually does it (refs kept, not inlined), and got 1,467,133 bytes across 328 endpoints, before the OData params we stick on top. So yeah, it's size.

Thanks for not just agreeing with me about the count thing.

One problem with the PR as it is now: in HTTP mode we build the server per request (`server.ts:757`, stateless), so registerGraphTools runs on every POST. The measuring pass is ~450ms here, so every tool call pays it. Would have to happen once.

But. I keep coming back to us fixing a client problem inside our own schemas. `--discovery` gives 3 tools instead of 328 and the whole budget thing just goes away. I brought it up once and then let it drop when you answered the preset part, my fault. Am seriously considering making it the default now.

What I'd like to keep from this is the measurement. We have nothing that tells you we just sent 328 tools and 1.4MB of schema, and that number holds even if the 1MB turns out to be wrong or moves later. The pruning I'd rather have opt-in behind the env var, like you suggested a couple of times.

Some smaller stuff too (z.union isn't handled, so a bunch of the calendar and todo bodies skip straight to the last strategy), but that only matters if it stays on by default.

What do you think?

okapi-ca:
Thanks — agreed on all three, and the per-request cost was a real miss on my side. Pushed as a second commit (f9d4e8c) so the delta is reviewable against what you read; I'll squash before merge if you prefer a single commit. PR description updated to match.

**Once per process.** Fit results are memoised per tool and per configuration fingerprint (ration, account names, max pages), and the aggregate is logged once per surface instead of on every `POST /mcp`. On 335 tools the first request pays 74 ms unrationed / 188 ms rationed, every request after that 9 ms.

**Opt-in.** `MS365_MCP_TOOL_DEFINITION_BUDGET` unset means nothing is pruned — every tool registers as generated. Setting it to a byte count turns the ration on (`4096` is the value measured to keep a full org surface at ~66 %). README documents it. What stays on regardless is the measurement.

**The measurement, made exact.** It now serializes exactly as `tools/list` does — the SDK conversion, then `normalizeToolSchemaRefs` — instead of re-converting with every `$ref` inlined. A test pins it to the wire form byte for byte against a real `McpServer`, and the logged aggregate equals a `tools/list` probe to the byte, both modes. On this branch, client regenerated today:

| | Unrationed | Ration 4096 |
|---|---|---|
| Schema bytes served, 335 tools | 1,831,221 B (174.6 %) | 688,812 B (65.7 %) |
| Skipped by the alphabetical walk | 123 | 0 |
| Tools pruned | — | 82 |
| `create-sharepoint-list` | 84,167 B | 3,447 B |

One thing worth knowing for your own number: since #643 inlines refs back wherever a client might not resolve `$defs`, what `tools/list` actually carries is larger than a refs-kept estimate. Measured through the transport with the OData params included, the unrationed surface is 1.83 MB here. The aggregate line reflects the post-#643 form because it runs the same normalization.

**`z.union`.** Handled — alternatives are pruned at the same depth (choosing between them is not a step deeper), and while I was there `z.record` (values one level down) and `.default()` (wrapper rebuilt around the pruned type), which also passed through whole. For what it's worth, on this surface the gap did not send bodies to the last strategy before or after: the 82 pruned land between d4/desc2 and d1/desc1, none reaches d1/desc0 or d0, and the opaque-body warning never fires. So it cost nested fidelity on the union subtrees rather than field names — fixed regardless, since it only gets worse as bodies grow.

**On `--discovery` as the default.** Our operator view, for whatever weight it carries: for an org connector, a few hundred tools the model can call directly is what our users signed up for, and discovery adds a search round-trip per task. It is a fine remedy and the over-budget warning now names it alongside the ration, `--enabled-tools` and splitting connectors. Whether it should be the default for everyone is your call; either way, with this PR an operator sees the number that decides whether they need any of them.

eirikb:
Hi. I've been thinking about this PR a lot. I understand it's a real issue, but perhaps only for Anthropic and it's possible to get around. I understand though the error might get obscured, so your fix is solid then.  
Since it is now opt-in, except for the warning but it's a warning only I think we should merge it :slightly_smiling_face: 

Two small things first though:

`fitToolDefinitionToBudget` is called outside the per-tool try/catch. Nothing throws today, I ran all 332 through it, but if it ever does then the whole registration dies instead of just that one tool. The utility loop right below does it inside the try, so I bet it's just an oversight.

The over-budget warning also fires on stdio, where the path is unaffected as you say yourself. So an org-mode CLI user gets told their tools are being silently dropped when they aren't. Gate it on `httpMode`, or maybe just word it as "reported for hosted connectors"?

PR #656, opened 2026-09-01
fix: respect an explicitly provided Accept header

jshtk:
## Problem

An `Accept` header passed explicitly by the caller is silently discarded for every endpoint that
declares `acceptType` in `endpoints.json`.

`Header` parameters are copied into `headers` at `src/graph-tools.ts:1506`, but a few lines later the
configured `acceptType` overwrites the entry unconditionally:

```ts
if (config?.acceptType) {
  headers['Accept'] = config.acceptType;
}
```

This makes the alternate representation of a resource unreachable through the tool, even when the
Graph API explicitly asks the client to retry with a different `Accept` value.

## Concrete failure

`get-meeting-transcript-content` is configured with `acceptType: 'text/vtt'`. When a tenant has
speaker attribution disabled (`EnableAttributedTranscripts = false`), Graph rejects the
speaker-attributed format and names the format to retry with:

```json
{
  "error": {
    "code": "Forbidden",
    "message": "Speaker-attributed transcript content is disabled for this tenant.",
    "innerError": {
      "code": "SpeakerAttributionNotAllowed",
      "message": "Speaker-attributed transcript content is disabled for this tenant. Retry with Accept 'application/vnd.microsoft.graph.transcript+text'."
    }
  }
}
```

Retrying with `Accept: application/vnd.microsoft.graph.transcript+text` returns the identical 403,
because the header never leaves the process — `text/vtt` is put back in its place. The transcript is
unreachable for that tenant, with no way for the caller to work around it.

Per the [Get callTranscript docs](https://learn.microsoft.com/graph/api/calltranscript-get), the
unattributed format is exactly the documented fallback here: "The speaker-attribution setting never
blocks the unattributed format."

## Change

Treat `acceptType` as a default rather than an override:

```ts
if (config?.acceptType && !headers['Accept']) {
```

Callers who send nothing keep the previous behaviour; callers who send an `Accept` header now get
what they asked for. This mirrors how a default should compose with an explicit argument, and keeps
the fix scoped to one condition.

## Tests

`test/accept-header-override.test.ts` covers both directions: the configured `acceptType` is still
applied when no header is supplied, and an explicitly supplied header survives.

## Notes

Found while retrieving Teams meeting transcripts on a tenant with speaker attribution turned off.
The same class of bug applies to any endpoint declaring `acceptType`, not just transcripts.

### DIFF
--- a/src/graph-tools.ts
+++ b/src/graph-tools.ts
@@ -52,6 +52,7 @@ import {
   CONFIRM_PARAM_DESCRIPTION,
   TIMEZONE_PARAM_DESCRIPTION,
   EXPAND_EXTENDED_PROPERTIES_PARAM_DESCRIPTION,
+  getAcceptParamDescription,
   getAccountParamDescription,
   getFetchAllPagesParamDescription,
 } from './lib/param-descriptions.js';
@@ -1524,6 +1525,11 @@ async function executeGraphTool(
           .replace(`{${camelCaseParamName}}`, encodedValue)
           .replace(`:${camelCaseParamName}`, encodedValue);
         logger.info(`Path param fallback: replaced :${camelCaseParamName} with encoded value`);
+      } else if (paramName.toLowerCase() === 'accept' && config?.acceptType) {
+        // The synthetic Accept param added for acceptType endpoints. It has no entry in
+        // the generated client's parameter list, so it lands here rather than in the
+        // 'Header' case above.
+        headers['Accept'] = `${paramValue}`;
       } else if (isOdataParam) {
         // Fallback: OData param recognised by name but absent from generated client's parameter
         // list — forward it as a query param rather than silently dropping it.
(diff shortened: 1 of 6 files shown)


eirikb:
Hi there, thanks for digging into this. The bug is real, docs say exactly what you say they do.

But I think the fix might be dead code 😬. No endpoint declares an `Accept` header param, so `headers['Accept']` is always empty at that line. The test only passes because the mock adds one. I think the param has to be exposed in the schema too, then your guard does its job.

What do you think?

jshtk:
You're right, thanks — the guard alone was dead code. No endpoint declares an `Accept` header param, so `headers['Accept']` was always empty there; the test only passed because the mock invented one.

Pushed a follow-up: tools whose endpoint declares `acceptType` now get a synthetic optional `Accept` param (same pattern as `fetchAllPages`/`timezone`), forwarded in `executeGraphTool`. `describeToolSchema` mirrors it so `--discovery` doesn't drift.

The mock no longer fabricates the param, and I added end-to-end tests against the real generated client — both `acceptType` tools apply their default and forward an explicit `Accept` into `graphRequest` — so a regression back to dead code can't hide behind a mock again.

eirikb:

PR #657, opened 2026-09-01
fix: correct and extend Graph query hints for chats and meetings

jshtk:
## Problem

Retrieving a Teams meeting transcript through this server took **13 tool calls, 7 of which failed
with HTTP 400**. Every failure was a query-option restriction the model had to discover by trial and
error — and one of them was actively caused by an incorrect `llmTip`.

Captured from a real session (Graph error messages quoted verbatim):

| Tool | Sent | Graph response |
| --- | --- | --- |
| `get-calendar-view` | `$select=…,joinWebUrl,…` | `Could not find a property named 'joinWebUrl' on type 'Microsoft.OutlookServices.Event'` |
| `list-online-meetings` | `$top`/`$orderby`/`$select`, no filter | `One of the required parameters to lookup meeting by QueryOptions is null or empty` |
| `list-online-meetings` | `$search="…"` | same `InvalidArgument` |
| `list-chats` | `$search="…"` | `Query option 'Search' is not allowed` |
| `list-chats` | `$orderby=lastUpdatedDateTime desc` | `QueryOptions to order by 'lastUpdatedDateTime' is not supported` |
| `list-chat-messages` | `$top=100` | `The limit of '50' for Top query has been exceeded` |
| `list-chat-messages` | `$orderby=createdDateTime asc` | `QueryOptions to order by 'CreatedDateTime' in 'Ascending' direction is not supported` |

### The incorrect hint

`get-calendar-view` and `get-specific-calendar-view` both ended with:

> Teams meetings include a **joinWebUrl** property needed for transcript access via
> list-online-meetings.

The event resource has no such property. Per
[Choose an API to create and join online meetings](https://learn.microsoft.com/graph/choose-online-meeting-api#comparing-the-apis),
the two APIs deliberately differ:

- Calendar API (`event`): "Through the **onlineMeeting** property, access **joinUrl**"
- Cloud communications API (`onlineMeeting`): "Use the **joinWebUrl** property"

A caller following the hint selects `joinWebUrl` on an event and gets a 400 — on the exact path the
hint recommends for reaching transcripts.

## Change

Four `llmTip` corrections, no code changes:

- **`get-calendar-view`, `get-specific-calendar-view`** — point at `onlineMeeting/joinUrl` and note
  that the event has no `joinWebUrl`, then show how to pass that value on.
- **`list-online-meetings`** — state that the `JoinWebUrl` filter is *required*, that the collection
  cannot be listed or paged, and that `$search`/`$orderby` are unsupported.
- **`list-chats`** (new) — no `$search`, no `$orderby` on `lastUpdatedDateTime`; match the topic
  client-side.
- **`list-chat-messages`** (new) — `$top` capped at 50, descending-only `$orderby`; meeting chats
  hold system events, transcripts come from `list-meeting-transcripts`.

Each restriction is taken from the Graph error it produces, so the hints stay checkable rather than
speculative.

## Tests

`npx vitest run` — 890 tests across 66 files pass, including `endpoints-validation`, `presets` and
`scope-derivation`.

Note: `format:check` is already failing on 132 files on `main`, `src/endpoints.json` among them, so
this change is formatted to match the surrounding entries rather than reformatted.

### DIFF
--- a/src/endpoints.json
+++ b/src/endpoints.json
@@ -510,7 +510,7 @@
     "scopes": ["Calendars.Read"],
     "supportsTimezone": true,
     "supportsExpandExtendedProperties": true,
-    "llmTip": "Returns expanded recurring event instances (not just seriesMaster) within a date range for the default calendar. Requires startDateTime and endDateTime query parameters in ISO 8601 format (e.g., 2024-01-01T00:00:00Z). Use get-specific-calendar-view if you need a non-default calendar. To find Teams meetings, use $filter=isOnlineMeeting eq true. To search by subject, use $filter=contains(subject,'keyword'). Teams meetings include a joinWebUrl property needed for transcript access via list-online-meetings."
+    "llmTip": "Returns expanded recurring event instances (not just seriesMaster) within a date range for the default calendar. Requires startDateTime and endDateTime query parameters in ISO 8601 format (e.g., 2024-01-01T00:00:00Z). Use get-specific-calendar-view if you need a non-default calendar. To find Teams meetings, use $filter=isOnlineMeeting eq true. To search by subject, use $filter=contains(subject,'keyword'). Teams meetings expose their join link as onlineMeeting/joinUrl (select it with $select=onlineMeeting); the event resource has no joinWebUrl property. Pass that joinUrl to list-online-meetings as $filter=JoinWebUrl eq '{url}' to reach transcripts."
   },
   {
     "pathPattern": "/me/calendars/{calendar-id}/calendarView",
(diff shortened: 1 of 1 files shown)


eirikb:
Hi, and thanks! The joinWebUrl/joinUrl part is spot on, that old tip was just wrong.

One thing on `list-chats`: `$orderby=lastMessagePreview/createdDateTime desc` is actually supported (there's an example in the docs), so I'd rather point at that than at `$select` + client-side matching. `$select` isn't in the supported list for `/me/chats` AFAIK. Could you tweak that one?

jshtk:
Good catch, thanks — you're right on both counts. The docs list $orderby as supporting `lastMessagePreview/createdDateTime` descending, and $select isn't among the supported options for `/me/chats`. I'd generalised from a single failing call (`$orderby=lastUpdatedDateTime`) to the whole option, which was too broad.

Updated the hint to name the supported options explicitly, including the $top max of 50. The other three tips are unchanged.

eirikb:

PR #668, opened 2026-09-09
feat: Add support for the skiptoken parameter

joelw:
This PR makes it possible for callers to send Graph's skiptoken parameter, which was previously filtered out.

The fetchAllPages parameter can take a long time if there are many pages. This change allows the caller to choose to follow nextLink pagination links on its own. This can be a bit more efficient because it can stop as soon as it finds the desired result - for example when trying to find a oneOnOne chat with a specific person using list-chats.

We make an attempt to only add this parameter to endpoints that probably allow it, rather than every single endpoint.

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -1089,6 +1089,143 @@ describe('graph-tools', () => {
     });
   });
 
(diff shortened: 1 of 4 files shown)


eirikb:

PR #671, opened 2026-09-11
feat(endpoints): let drive item tools run under Sites.Selected

cbjerg:
## Problem

The README documents `Sites.Selected` as a supported SharePoint permission model with `--allowed-scopes`, and the `/sites/...` tools already carry `Sites.Selected` as an OR-group alternative. The `/drives/{drive-id}/...` item tools do not: they declare only `Files.Read` / `Files.ReadWrite`, so an allowlist like `--allowed-scopes 'User.Read Sites.Selected'` hides every file operation:

```
disabledTools: list-folder-files, upload-file-content, create-upload-session, get-drive-item,
               move-rename-onedrive-item, create-onedrive-folder, copy-drive-item, list-drive-item-versions
missingAllowedScopesForTools: Files.Read, Files.ReadWrite
```

That leaves a selected-site deployment able to list a site's drives but unable to read or write anything in them — while Graph itself serves `/drives/{id}/items/...` for any drive of a granted site collection.

## Change

Nine drive item tools become OR-groups with `Sites.Selected` as the second alternative, the same shape `get-sharepoint-site` etc. already use:

```json
"scopes": [["Files.ReadWrite"], ["Sites.Selected"]]
```

The primary group is unchanged, so login scopes without an allowlist are identical (`--list-permissions` still reports `Files.ReadWrite`). `scopes` is kept (not `workScopes`) so personal-account behaviour is untouched. No code change.

Also: one README sentence, and `test/sites-selected-drive-tools.test.ts` (fails on the previous `endpoints.json`, passes now; asserts enabled-under-Sites.Selected, disabled-without-either, and unchanged login scopes).

## Verification

- `--list-permissions` with the allowlist: `disabledTools: []`, `effectivePermissions: ['Sites.Selected']`.
- End-to-end against a tenant where the app is granted `write` on one site and `read` on another (delegated `Sites.Selected` only, admin-consented), signed in as a user who has Edit on both in the browser: upload to the read-only site → 403 `accessDenied`; upload to the write site → 201; listing an ungranted site's library → 403; listing the read site's library → 200.
- `npm run typecheck`, `lint`, `format:check` clean. `npm test`: the only failures on my machine are the 5 in `test/attachment-split-listener.test.ts`, which fail identically on `main` here (port binding), unrelated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ SharePoint supports two enterprise permission models:
 - Broad tenant scopes such as `Sites.Read.All`, `Sites.ReadWrite.All`, and `Sites.Manage.All`.
 - Microsoft Graph `Sites.Selected`, where SharePoint site access is granted to the app on specific site collections and Graph evaluates the signed-in user's own permissions at request time.
 
-The default org-mode behavior continues to request the broad SharePoint scopes used by existing deployments. Enterprises that want selected-site SharePoint access can set an allowlist containing `Sites.Selected` instead of broad `Sites.*.All` scopes. Direct site/list/item tools that target an explicit SharePoint site can run with `Sites.Selected`; tenant-wide SharePoint discovery and search tools still require broad SharePoint scopes.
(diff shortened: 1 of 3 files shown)


eirikb:

PR #672, opened 2026-09-11
feat(http): stateless two-leg PKCE for multi-instance hosts (MS365_MCP_PKCE_SECRET), plus noListen

cbjerg:
## Problem

The two-leg PKCE mapping (`pkceStore`) is process memory keyed by OAuth `state`, so `/authorize` and `/token` must hit the same process. `docs/deployment.md` recommends `--max-replicas 3` on Container Apps, and any serverless host has the same shape: when `/token` lands on another instance there is no mapping, the client's verifier is forwarded upstream against the server's challenge, and Entra answers `AADSTS501481`. Sign-in then fails intermittently, which is the worst kind of failure to debug.

## Change

Opt-in, no behaviour change without it. With `MS365_MCP_PKCE_SECRET` (32+ chars) set:

- `/authorize` derives the server-side verifier as `HMAC-SHA256(secret, client code_challenge)`, sends `sha256(verifier)` upstream, stores nothing.
- `/token` recomputes it from `sha256(client code_verifier)` and uses it for the upstream exchange.

An attacker who sees the challenge in the redirect cannot reproduce the verifier without the secret — the same guarantee the store gave. A secret shorter than 32 chars is refused at startup rather than silently weakening the derivation. Without the env var the existing store path is untouched.

Also `options.noListen` (not a CLI flag): `start()` builds the Express app and exposes it via `getHttpApp()` instead of binding a port, so an embedding host can mount it. One paragraph in `docs/deployment.md` for both.

## Verification

- `test/stateless-pkce.test.ts`: derivation is deterministic/secret-dependent/RFC 7636-shaped; secret length enforced; and the multi-instance case — `/authorize` on one `MicrosoftGraphServer`, `/token` on a *fresh* one — sends the server verifier with the secret and the client verifier without it (the documented failure).
- In production on Vercel (one function, many instances): dynamic client registration → `/authorize` → real Microsoft sign-in → `/token` → bearer → Graph tool calls, all succeeding. The same deployment without the secret cannot complete sign-in reliably.
- `npm run typecheck`, `lint`, `format:check` clean; OAuth-related suites pass. `test/attachment-split-listener.test.ts` fails identically on `main` here (port binding on this machine), unrelated.

Independent of #671 (Sites.Selected on drive tools); either can land first.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -206,6 +206,7 @@ The client automatically discovers OAuth endpoints and opens a browser for authe
 ## Security Considerations
 
 - **Stateless**: the server does not store tokens — each request carries the user's Bearer token
+- **More than one instance? Set `MS365_MCP_PKCE_SECRET`** (32+ random characters). The two-leg PKCE flow otherwise keeps a mapping in process memory between `/authorize` and `/token`, so a second replica, a serverless function or a restart between the two legs fails the exchange with `AADSTS501481`. With the secret set, the server-side verifier is derived from the client's code_challenge and the secret and nothing is stored, so any instance can complete the exchange. Single-process deployments can leave it unset. Embedding hosts can also build the app without binding a port via `options.noListen` + `getHttpApp()`.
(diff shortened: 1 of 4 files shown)


eirikb:
Hi, LGTM, but can you fix the build? Thanks :slightly_smiling_face:

PR #674, opened 2026-09-12
fix(tools): tighten skiptoken cursor handling and shorten its description

eirikb:
Follow-up to #668.

- A pasted `@odata.nextLink` that pages with `$skip` (Outlook mail and calendar) now sends `$skip`, instead of forwarding the whole URL as a cursor.
- A link carrying no cursor, such as a drive or sites delta `token=` link, is refused with a clear error instead of a garbage request.
- `skiptoken: ""` no longer goes out as an empty `$skiptoken=`.
- The cursor is matched only at the start of the value or after `?` or `&`, so a cursor-looking value inside another query param cannot be read as the cursor.
- The parameter description is one sentence now. It rides on ~101 tools, where the old text cost about 40 KB of tool schema.

### DIFF
--- a/src/__tests__/graph-tools.test.ts
+++ b/src/__tests__/graph-tools.test.ts
@@ -1292,14 +1292,20 @@ describe('graph-tools', () => {
       else process.env.MS365_MCP_ALLOW_PAGINATION = prevAllowPagination;
     });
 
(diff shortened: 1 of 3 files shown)


joelw:
@eirikb Thanks so much for improving my original PR!

PR #676, opened 2026-09-14
fix: share query parameter contracts across discovery and execution

gnegno84:
## Problem

Query schemas differ between normal MCP registration and discovery mode. `describeToolSchema` uses generated parameter types, while `registerGraphTools` separately replaces several types. For example, a caller can discover an array for `select` and then encounter a string-only contract in normal registration.

The generated collection parameters also disagree with documented Graph endpoint constraints:

- `list-joined-teams` and `list-my-associated-teams` advertise unsupported OData options. Their collection parameters also cause synthetic `skiptoken` injection.
- `list-chats` does not enforce the documented maximum page size of 50 in its input schema.
- Passthrough inputs and `execute-tool` can reach execution without enforcing the corrected query contract.

## Change

Use one shared `queryParameterSchema` function for registration, discovery, and execution, replacing the separate registration-only type overrides.

- Accept strings or string arrays for the comma-separated `select`, `expand`, and `orderby` options. Existing serialization sends both forms as comma-separated text. No extra Graph request or retry is added.
- Exclude unsupported query options from both team-list schemas. Without those collection parameters, neither path injects `skiptoken`. Reject stale unsupported OData arguments before Graph dispatch, including dollar-prefixed names.
- Limit chat `top` to 50 while retaining any stricter provider constraint. Reject an oversized page rather than silently reducing the requested result count.
- Preserve existing calendar-delta handling for stale `top` inputs and keep `fetchAllPages` available.
- Check effective query schemas at execution so the passthrough/discovery route follows the same contract.

This corrects the connector's definitions and execution boundary. It does not infer user intent, silently drop requested filters, automatically fetch additional pages, or change OAuth behavior.

## Tests

- Compare query schema types, constraints, and requiredness between registration and discovery across the generated tool registry.
- Verify team-list options and synthetic cursors are absent from both schema paths, while `fetchAllPages` remains available.
- Verify chat page limits, including a stricter original bound.
- Verify array field selection produces one Graph request with comma-separated text.
- Verify unsupported team options and oversized chat pages fail before Graph dispatch.

## References and related PRs

- [List joinedTeams: unsupported OData options](https://learn.microsoft.com/en-us/graph/api/user-list-joinedteams?view=graph-rest-1.0)
- [List associatedTeamInfo: unsupported OData options](https://learn.microsoft.com/en-us/graph/api/associatedteaminfo-list?view=graph-rest-1.0)
- [List chats: supported options and page limit](https://learn.microsoft.com/en-us/graph/api/chat-list?view=graph-rest-1.0)

The two open PRs were checked before preparing this change. #672 concerns OAuth. #655 concerns oversized request-body definitions; it touches nearby registration code but does not address query type consistency or these endpoint constraints. The changes should remain separate, with registration edits reconciled if either PR lands first.


eirikb:

Issue #546, opened 2026-06-18
`download-bytes` returns JSON-parseable content as a re-serialized object (lossy), not bytes


ryant-calterra:
### Summary
`download-bytes` advertises base64 bytes (*"Returns { contentType, encoding:"base64", contentBytes }"*)
but actually returns three shapes by response body — and the JSON case is lossy:

| Response body | Returns | Faithful? |
|---|---|---|
| binary content-type (pdf, image, octet-stream, office) | `{ contentType, encoding:"base64", contentBytes }` | ✅ |
| non-binary, doesn't parse as JSON (`.md`, `.css`, `.svg`, plain text) | `{ message:"OK!", rawResponse:"<text>" }` | ✅ |
| non-binary, parses as JSON (a `.json` item; `application/json` is non-binary) | the parsed value, no envelope, re-serialized | ❌ |

For the JSON case the bytes go through `JSON.parse`→`JSON.stringify`, so the result isn't
byte-identical (whitespace/indentation collapsed, trailing newline dropped, duplicate-key order and
number formatting not preserved) and the shape itself changes to a bare value. The trigger is
**parse-success, not content-type** — `JSON.parse` is tried on *every* non-binary body, so a
`.txt`/`.csv` whose whole content is valid JSON is also affected, while a malformed `.json` comes
back faithfully as `rawResponse`.

**Affected version:** `v0.114.5`; unchanged on `main` (`3537ce0`).

### Root cause
1. `/content` forces `rawResponse=true` (`src/graph-tools.ts:655-665`).
2. `makeRequest` (`src/graph-client.ts:142-153`) — for non-binary bodies, `JSON.parse(text)` is
   returned directly as the result; only a parse *failure* yields `{ message, rawResponse: text }`.
   `application/json` isn't binary (`isBinaryContentType`, `:19-47`), so `.json` takes this branch.
3. `rawResponse=true` re-serializes via `JSON.stringify` (`formatJsonResponse :297-300` →
   `serializeData :202-211`) — the lossy round-trip.

### Repro
Download a pretty-printed `.json` drive item's `/content`; it comes back minified/reordered, while a
sibling `.md`/`.pdf` round-trips faithfully.

### Suggested fix
`rawResponse=true` is set in exactly one place (`src/graph-tools.ts:665`), so skipping the parse for
it doesn't affect normal Graph calls (those run `rawResponse=false` and need JSON parsing). In
`makeRequest`'s non-binary branch:

```ts
[...]

PR #679, opened 2026-09-15
fix(graph-client): `download-bytes` returns base64 for every content type; the client never returns lossy text

### LINKED ISSUE #678: `download-bytes` corrupts attachments whose Content-Type is not on the binary allowlist (e.g. `application/msword`)
**What happens (0.154.0).** `download-bytes` on a Word 97-2003 mail attachment (`application/msword`) returns `{ message, rawResponse }` with no `contentBytes`. `rawResponse` is the body decoded as UTF-8, so every invalid byte sequence has become U+FFFD — 6,152 of them in a 49,664-byte file — and the file cannot be rebuilt. The same attachment fetched through the attachment JSON (`list-mail-attachments` → `contentBytes`) decodes byte-for-byte to the original, so the bytes are fine on the wire; the loss is in the client.

**Why.** `graphRequest` reads the body with `arrayBuffer()` only when `isBinaryContentType(Content-Type)` is true: `image/*`, `video/*`, `audio/*`, `font/*`, `application/octet-stream`, `application/pdf`, `application/zip`, `application/vnd.*` and `application/x-*` (minus `+json`/`+xml`/`+text`). Every other type goes through `response.text()`. `download-bytes` calls `graphRequest(target, { rawResponse: true })` and never asks for binary handling, although its documented contract is `{ contentType, encoding: "base64", contentLength, contentBytes }`. This is the same class of bug as #546, for non-JSON bodies.

**Probably also affected** (only `application/msword` confirmed): `application/rtf`, `application/ms-tnef` (winmail.dat), `application/pkcs7-signature` / `pkcs7-mime` (S/MIME), `application/postscript`, `application/gzip`, `message/rfc822` (.eml with 8-bit content), and any `text/*` attachment in a non-UTF-8 encoding.

**Impact.** Over stdio, `download-bytes-to-file` streams every type correctly and is a workaround. In `--http` mode that tool does not exist, so a hosted MCP client has no way to get these bytes at all.

**Reproduce.**
1. `list-mail-attachments` on a message carrying a `.doc`; note the attachment id and `contentType: application/msword`.
2. `download-bytes` with `target=/me/messages/{id}/attachments/{attId}/$value`.
3. The result has `rawResponse` and no `contentBytes`; `rawResponse` contains U+FFFD. A `.pdf` or `.docx` on the same message returns `contentBytes` as expected.

**Proposed fix.** A `forceBinary` option on `graphRequest`, checked before the Content-Type allowlist, and set by `download-bytes` — the tool's contract is the bytes, whatever Graph reports. Widening the allowlist would only move the failure to the next unlisted type. PR: #679.

Environment: `@softeria/ms-365-mcp-server` 0.154.0, Node 22.23, macOS; personal account; reproduced over stdio and over `--http --trust-proxy-auth`.

JEM-Fizbit:
Fixes #678.

**Root cause.** `graphRequest` reads the body with `arrayBuffer()` only when `isBinaryContentType(Content-Type)` matches its allowlist; everything else goes through `response.text()`, which replaces invalid UTF-8 with U+FFFD and drops a leading byte-order mark. `download-bytes` passed `rawResponse: true` but never asked for binary handling, so an `application/msword` attachment (and `application/rtf`, `message/rfc822`, a non-UTF-8 `text/*`, …) came back as lossy `rawResponse` text with no `contentBytes`. Same class as #546, for non-JSON bodies.

Two commits, reviewable separately:

**1. `download-bytes` asks for bytes** — `GraphRequestOptions.forceBinary`, checked before the allowlist; `download-bytes` sets it. The tool now returns its documented `{ contentType, encoding: "base64", contentLength, contentBytes }` for every type. A JSON body is base64 too (byte-faithful, which is what #546 asked for) and an empty body yields `contentLength: 0`; `rawResponse: true` is kept for the #546 path but is not reached while `forceBinary` is set. **Callers that read `rawResponse` from `download-bytes` for JSON or text bodies must switch to decoding `contentBytes`** — that field was never the documented shape, but it is a visible change.

**2. The invariant, for every caller** — read every body as bytes; hand it on as text only when the caller did not ask for bytes, the type is not on the binary allowlist, *and* it decodes as strict UTF-8 (`TextDecoder('utf-8', { fatal: true, ignoreBOM: true })`); otherwise return the base64 shape. A Content-Type we failed to list, or a `text/*` body in another encoding, can no longer come back with its bytes replaced. Graph's JSON is always UTF-8, so JSON endpoints are unchanged apart from tolerating a BOM before `JSON.parse`; `response_bytes` in the audit metadata is now the true byte count rather than the length of the decoded text. Cost: one buffer copy per text response.

No change to `download-bytes-to-file` (already streams) or to any endpoint tool's request path.

**Tests.** `graph-client.test.ts`: `forceBinary` round-trips an OLE2-header + invalid-UTF-8 body byte-for-byte; a table over `application/msword`, `application/rtf`, `message/rfc822` and Windows-1252 `text/plain` proves the fallback without `forceBinary`; valid UTF-8 text still returns as text with its BOM intact; JSON with a BOM still parses. `graph-tools.test.ts`: `download-bytes` passes `forceBinary` and `rawResponse`. Three existing tests (`http-account-param`, `http-oauth-fix`, `request-context`) faked a fetch Response with only `text()`; they now also provide `arrayBuffer()`.

**Verification.** `npm run lint` (0 errors), `format:check`, `typecheck`, `build` clean. `npm test`: 1435 passed; the 5 failures in `test/attachment-split-listener.test.ts` are pre-existing on this machine (they bind 127.0.0.2, which macOS does not configure) and fail identically on `main`. Live: the patched build over stdio fetched a 49,664-byte `.doc` attachment whose sha256 matches the bytes obtained independently from the attachment JSON, where 0.154.0 returns 6,152 replacement characters.


eirikb:

PR #684, opened 2026-09-21
feat: MS365_MCP_GRAPH_BASE_URL overrides the Graph API base URL

jo3zik:
## What

Adds `MS365_MCP_GRAPH_BASE_URL`: an environment variable that sends every Graph call to a base URL of your own instead of the cloud's `https://graph.microsoft.com`.

## Why

Deployments that put an egress proxy in front of Graph. Today the server proxies only over CONNECT (`HTTPS_PROXY`), which is a TLS tunnel: a proxy that wants to add the bearer token itself, or route by tenant, cannot see the request without a CA installed in the container. With the base URL pointed at the proxy, the server dials it over plain HTTP and the proxy originates TLS to Microsoft.

## How

- `getCloudEndpoints()` is the one place every Graph URL is built from (`graph-client.ts`, the `/me` probes in `auth.ts` and `oauth-provider.ts`, `obo-client.ts`, the discovery route in `server.ts`), so the override lands there and replaces `graphApi` for whichever cloud is selected. The login authority is untouched.
- The value must be an absolute http(s) URL. A path prefix is kept (`http://proxy:10255/tenant-a/graph` + `/v1.0/me`); a trailing slash is dropped. Blank is unset; anything else throws with the variable's name.
- Deliberately **not** added to the `.env` allowlist in `load-env.ts`: a file in the client's cwd must not be able to redirect bearer-token traffic (same reasoning as GHSA-9w34-3f56-vwmh).
- README: one line in the environment variable list and a pointer under Supported Clouds.

## Tests

`test/graph-base-url.test.ts`: unset leaves both clouds alone; set replaces `graphApi` for both and keeps `authority`; trailing slash stripped; blank ignored; non-URL rejected; and a `GraphClient.graphRequest('/me/messages')` with the variable set hits `http://127.0.0.1:10255/tenant-a/outlook/v1.0/me/messages` (fetch mocked).

`npm run verify` is green locally (generate, lint, format:check, typecheck, build, test: 1492 passed).


eirikb:
Hi there

Thanks for this, and especially for keeping it off the `.env` allowlist because of GHSA-9w34-3f56-vwmh thing. Nice that you found that one.

Two things before I merge:

1. `obo-client.ts` uses `graphApi` as the OAuth resource, not as a URL, so with the variable set it asks Entra for `http://127.0.0.1:10255/tenant-a/outlook/.default` and OBO fails with AADSTS500011. I think overriding the base at the `graph-client.ts` call site is safer than replacing `graphApi` in `getCloudEndpoints()`.

2. The README says "every Graph call", buuut `get-download-url` returns Graph's pre-authenticated SharePoint URL, and the 302 from `/content` goes straight to SharePoint as well. For someone doing egress control that difference matters, could you mention it?

Rest looks fine to me. Also, I usually like to run PRs before merging and I don't have a setup for this one - does it work for you in production? 🙂

